PR #18, opened 2022-05-03
Stub form api

0atman:
Create stubs:
- [x] Create a form using the API
- [x] Read a form
- [x] Update a form
- [x] Delete a form (optional)
- [x] List forms (all forms, optional)

If we quickly publish a stub API, we can work to the same interface while building out functionality.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


0atman:
@danielburnley Just checking I've understood the layout we have Services, which are controllers, and Repositories, which are Models, and the Server routes act as Views?

danielburnley:
@0atman Sorta yea - though this isn't really MVC as such, more split by responsibilities

- `server.rb` for now contains the actual routes, which is where the logic for kinda processing the HTML requests will go (see: turning request bodies into ruby hashes) 
- `services` will handle the actual business logic, these shouldn't know _anything_ about HTML stuff and will operate on just ruby hashes - Unit tests for this can therefore not need to do anything db related for setup and can stub repositories via dependency inversion stuff
- `repositories` will be where anything database related can go, similarly to the `server.rb` anything related to databases should stay here and services shouldn't see any

At least, eventually that'll be the case. In the short term we arguably don't need the service layer if all we're doing is parsing a request and throwing it straght into the DB, kinda becomes a bit like a pointless layer in that case. 

For stubbing the API however, we can probs ignore all of that and just return the json in the `server.rb` straight away without the need for any other layers

0atman:
Swagger inspector looks good (run the server locally and try it here https://inspector.swagger.io/builder?url=http%3A%2F%2Flocalhost%3A9292%2Fapi%2Fv1%2Fswagger_doc)

Screenshot from 2022-05-09 11-33-32

lib/server.rb
```diff
@@ -1,10 +1,29 @@
...
+  resource :forms do
+    desc 'Returns all forms.'
+    params do
+      requires :result, type: Integer, desc: 'Result ID.'
+    end
+    route_param :result do
+      get do
+        Services::Example.new.execute(params[:result])
+      end
+    end
```
0atman, line 18: @danielburnley calling this out early, like the format of grape so far? Big example looks pretty tasty to me https://github.com/ruby-grape/grape#basic-usage

0atman:
So far

0atman:
Tweaked schema based on our chat earlier
Screenshot from 2022-05-10 14-44-41

lib/server.rb
```diff
@@ -1,10 +1,67 @@
...
+  resource :forms do
+    desc "Create a form."
+    params do
+      requires :form, type: String, desc: "Form data."
+    end
+    post do
+      {
+        form_id: 1,
+        form_name: "form name",
+        user: "user@example.com"
```
danielburnley, line 18: i'd probably go with:

- `id` (the form is redundant)
- `name` (same)
- `submission_email` - it'll be where we send the submissions

0atman: Nice, yeah that makes sense

0atman:
Format now

```json
{
    "id": 12,
    "name": "form name",
    "submission_email": "user@example.com"
}
```

lib/server.rb
```diff
@@ -1,10 +1,67 @@
...
+          name: "form name",
+          submission_email: "user@example.com"
+        }
+      end
+    end
+
+    desc "Update a form."
+    params do
+      requires :id, type: String, desc: "Form ID."
+      requires :data, type: String, desc: "Form data."
```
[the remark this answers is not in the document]
0atman: cheers, fixed

lib/services/example.rb
```diff
@@ -2,4 +2,14 @@ class Services::Example
...
+
+  def get_forms(name)
+    JSON({
+           name => {
+             item2: {
+               item3: "ok"
+             }
+           }
+         })
+  end
```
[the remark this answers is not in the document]
0atman: gonna be kind to github's storage and remove it, well spotted!

danielburnley (approved):

0atman:
Listing forms OK:

Screenshot from 2022-05-11 11-45-37

PR #20, opened 2022-05-11
changes badges to native github

0atman:


0atman:
I should have used native github badges. woops.

PR #22, opened 2022-05-18
Add docker dB

0atman:
for convienience, and maybe we'll set up the whole stack in docker at some point?

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,13 @@
+version: "3.9"
+services:
+  db:
+    image: postgres
+    environment:
+      POSTGRES_USERNAME: postgres
+      POSTGRES_PASSWORD: password
+    ports:
+      - "5432:5432"
+    volumes:
+      - forms-api-pgdata:/var/lib/postgresql/data/
+volumes:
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,13 @@
+version: "3.9"
+services:
+  db:
+    image: postgres
+    environment:
+      POSTGRES_USERNAME: postgres
+      POSTGRES_PASSWORD: password
+    ports:
+      - "5432:5432"
```
danielburnley, line 9: do we want to make sure this maps differently on each of our repos? otherwise you can't run all three in docker at once as they'll all bind do the same port 🤔

danielburnley: this does end up making each initial set up a bit more of a pain tho, hrm

0atman: Ah yeah, I think we'd probably run the whole stack inside docker-compose - no need to expose any ports to the local host. 
This config just works for the forms-api locally, local ruby, docker db. Next step is the bigger one.

PR #29, opened 2022-06-16
Add start page and next page information

danielburnley:

The runner requires this information to load the initial page as well as to navigate between them.
Following this we should look to create tested service objects for this, however as I'm unsure if these are likely to change I've left them untested in the webroutes for now.

#### Checklist

### DIFF
--- a/lib/server.rb
+++ b/lib/server.rb
@@ -11,7 +11,10 @@ class Server < Grape::API
   end
 
   rescue_from :all do |e|
-    Sentry.capture_exception(e)
+    unless ENV['SENTRY_DSN'].nil?
+      Sentry.capture_exception(e)
+    end
+
     error! e.message, 500
   end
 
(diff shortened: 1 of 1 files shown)


lib/server.rb
```diff
@@ -51,7 +52,12 @@ class Server < Grape::API
       desc "Read a form."
       get do
         repository = Repositories::FormsRepository.new(@database)
-        repository.get(params[:form_id])
+        page_repository = Repositories::PagesRepository.new(@database)
+
+        form = repository.get(params[:form_id])
+        pages = page_repository.get_pages_in_form(params[:form_id]).sort_by { |page| page[:id] }
+        form[:start_page] = pages.first[:id]
```
thomasiles, line 59: I think this fails with a 500 when the form doesn't have any pages which causes the runner to fail too.

thomasiles (approved):

PR #36, opened 2022-06-29
add organisation to api

0atman:

The gds sso org should be saved into the form, so that only members of that organisation can edit/view it.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


spec/db/migration_4_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "pry"
```
alice-carr, line 1: we should remove pry

0atman: Yep, agreed.

alice-carr (approved):
looks good

Issue #32, opened 2022-06-21
Add an API key to the API


danielburnley:
Though we may change the implementation in the future, to begin with we just want to add a shared API key to the API that's sent in a header by the runner/admin

- [x] An API key is required to be sent in the header `X-API-Key` 
- [x] The API returns a 401 when the API key is not specified

API_KEY has been added as an Environment Secret to:
- [x] forms-api
- [x] forms-runner
- [x] forms-admin

PR #38, opened 2022-07-11
Api key

alice-carr:
relates #32 
This adds authentication to the api

#### Checklist

    - [x] README.md


README.md
```diff
@@ -29,7 +29,7 @@ git clone git@github.com:alphagov/forms-api.git
 cd forms-api
 
 # 2. Run the setup command
-make install
+make setup
```
alice-carr, line 32: I added a drive by fix for this mistake in the readme

spec/lib/server_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  include Rack::Test::Methods
+
+  def app
+    Server::Server
+  end
+
+  context "GET /api/v1/forms" do
+    it "returns a status code 401 when no authentication supplied" do
+      get "/api/v1/forms"
+      expect(last_response.status).to eq(401) unless ENV["API_KEY"].nil?
```
alice-carr, line 13: I was having trouble dealing with the case where the API_KEY env hasn't been set - I've just bypassed he first and last tests if the env hasn't been set. I feel like this isn't the best way to do this, so please let me know if you have any better ideas

alice-carr:
I'm going to mark this as ready to review now - we won't want to actually merge this until all places where the api is called has been updated to pass in the X-API-Key in the header. 
Note I've added API_KEY as a repository secret in this repo

spec/lib/server_spec.rb
0atman: hmmmmmmmmmm I think we should mock this in the tests?

alice-carr: yeah good point

alice-carr: ok I've stubbed the api key env var now.

lib/server.rb
```diff
@@ -6,6 +6,12 @@ class Server < Grape::API
   format :json
   prefix :api
 
+  helpers do
+    def authenticate
+      error!("Unauthorized", 401) unless headers["X-Api-Token"] == ENV["API_KEY"] || ENV["API_KEY"].nil? && !headers["X-Api-Token"].nil?
```
alice-carr, line 11: I've set this up so that if no api_key is supplied in the env vars, the api is essentially open - is this the way we want to approach this? Should we do the opposite - ie if no api_key is in the env vars, the api is locked down completely? Am I overthinking this since it's unlikely there will be a scenario where the env var is empty?

alice-carr:
I've started the work to pass in the api token in the runner and admin here: 

- https://github.com/alphagov/forms-runner/pull/55
- https://github.com/alphagov/forms-admin/pull/70

I've started trying to fix the tests but have been having a bit of a nightmare with the active resource mocks.

alice-carr:
I've also converted this back to a draft PR as I imagine we won't be able to work on this for a bit (seems other work will take precedence) and I wanted to keep the slack angry seal happy.

alice-carr:
Finally made headway with the tests, but there are three in forms-admin that are still giving me grief. 
All that's left to do before this is done is:
- [ ] fix those three tests
- [ ] one final manual test

DavidBiddle (approved):
I think this all looks good - there's the open question you raised of what should happen if the env variable is blank, but maybe the current behaviour is fine for now and we can change it later if we decide we don't like it?

lib/server.rb
aliuk2012: As discussed during this morning tech weekly. We should lock the API down by default and if we need to make it public (i.e in the future we might want to develop and api for organisations to use to import data directly to their own backend services, we would develop that later)

alice-carr: Just changed this so that if no api_key is supplied in the env vars the api is locked down

PR #39, opened 2022-07-12
tweaked forms list to require org

0atman:
The frontend requires filtering forms by org - this is the default behaviour.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


Makefile
```diff
@@ -23,4 +23,4 @@ lint-fix:
 db:
 	docker-compose up -d
 clean:
-	docker-compose down
\ No newline at end of file
+	docker-compose down
```
0atman, line 26: oh gods whose editor doesn't understand how unix lines work. I better not be vscode

0atman:
Filtering by get param:

Screenshot from 2022-07-12 13-21-44

spec/repositories/forms_repository_spec.rb
```diff
@@ -21,6 +21,13 @@
       expect(form[:submission_email]).to eq("submission_email")
       expect(form[:org]).to eq("org")
     end
+    it "gets a form by org" do
+      subject.create("name", "submission_email", "org")
```
danielburnley, line 25: Might be worth making a couple to make sure it returns two?

alice-carr (approved):
looks good!

spec/repositories/forms_repository_spec.rb
0atman: Very smart, added that

Makefile
DavidBiddle: I don't think vscode is the culprit - it's actually pretty good at picking up Editorconfig settings

DavidBiddle (approved):
Looks good  - presumably this will break the admin homepage until I've done my bit but that's fine I think

PR #43, opened 2022-07-26
Add new date fields to db and api

0atman:
#### Checklist

    - [ ] README.md
- [ ] set updated_at and created_at to the current time of the migration
- [ ] bump updated when the pages update/delete/create routes are hit


spec/db/migration_5_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+  it "adds a sensible default published from v4 to v5" do
+    form_id = database[:forms].insert(name: "name", submission_email: "submission_email", org: "testorg")
+
+    migrator.migrate_to(database, 5)
+
+    updated_form = database[:forms].where(id: form_id).first
+    expect(updated_form[:published]).to eq(false)
+  end
+end
```
0atman, line 17: @alice-carr just looks at this. I LOVE unit testing migrations omg

spec/db/migration_5_spec.rb
```diff
@@ -0,0 +1,17 @@
...
```
0atman, line 17: @alice-carr just looks at this. I LOVE unit testing migrations omg

0atman:
@thomasiles has suggested that 'published' be a datestamp, and we update it on form updated. I don't hate this idea, but maybe it should be called "updated"? But then how do we gate publishing/draft?

0atman:
Maybe add:
`published_at`
and
`updated_at`
?

DavidBiddle:
Agree that all 3 of those sound like sensible fields

0atman:
will do this today or tomorrow :ok_hand:

thomasiles:
Makes sense - I guess updated at will be changed when pages are updated too? Maybe we should go with some other wording than published_at now if it's easier?

DavidBiddle:
Feels like `made_live_at` would fit the wording we're using?

0atman:
`made_live_at` feels a little clunky, why not just `live_at`, especially as "made" is past tense, and we'll potentially be post-dating this?

alice-carr:
I think `published_at` reads better than `live_at` - I can imagine finding `live_at` hard to decipher if I saw it without context

0atman:
@alice-carr I think the problem is that the wider team have decided on some very specific meanings for _live_ and _published_, our abstraction is leaking

alice-carr:
@0atman yeah that makes total sense

DavidBiddle:
> `made_live_at` feels a little clunky, why not just `live_at`, especially as "made" is past tense, and we'll potentially be post-dating this?

I agree it's a bit clunky, but `created_at`, `published_at` and `updated_at` are all past tense as well so on that front it's consistent.

Are we content designing a DB field? 😄

0atman:
API design counts! Created, Updated, ARE past tense, though, these events will always be set to `Time.now`, whereas `live_at` can be set for the future, right?

0atman:
This is now fully working, I just tested the flow.

# After initial creation
```json
{
    "id": 1,
    "name": "form1",
    "submission_email": "tris@example.com",
    "org": "an-org3",
    "created_at": "2022-08-03 15:43:46 +0100",
    "live_at": null,
    "updated_at": null,
    "start_page": null
}
```

# Editing but no live date
```json
{
    "id": 1,
    "name": "form1",
    "submission_email": "tris@example.com",
    "org": "an-org3",
    "created_at": "2022-08-03 15:43:46 +0100",
    "live_at": null,
    "updated_at": "2022-08-03 15:45:46 +0100",
    "start_page": null
}
```

# Editing and adding a live date
```json
{
    "id": 1,
    "name": "form1",
    "submission_email": "tris@example.com",
    "org": "an-org3",
    "created_at": "2022-08-03 15:43:46 +0100",
    "live_at": "2022-08-13 11:11:11 +0100",
    "updated_at": "2022-08-03 16:09:52 +0100",
    "start_page": null
}
```

@DavidBiddle @thomasiles would love your review, as you'll be consuming this - does it make sense?

thomasiles:
Amazing! 

Two questions:
 - would it make sense to set `updated_at` and `created_at` to the current time of the migration? That's dirty but otherwise the possibility of them being empty has to be handled by clients and we don't have to worry about production (I don't think)
 - Would it make sense to bump updated when the pages update/delete/create routes are hit? Either that or add similar fields to the pages table?

0atman:
Aha! I'd forgotten about pages. From a user's perspective when a page is updated, the FORM is updated. Yes, i'll do that.

I see no harm with that migration suggestion - seems more coherent for current test data.

0atman:
Sorry I didn't check this in on Wednesday - this is ready! @thomasiles I've added tests to cover modifying a page updating the `updated_at` field as suggested :ok_hand:

db/migrations/4_add_org.rb
```diff
@@ -1,3 +1,5 @@
+require "pry"
+
```
aliuk2012, line 2:
```suggestion
```

db/migrations/5_add_published.rb
```diff
@@ -0,0 +1,17 @@
+require "pry"
+
```
aliuk2012, line 2:
```suggestion
```

aliuk2012:
Might be worth rebasing this branch with main branch (Code coverage should be enabled).

Apart from that looks really good.

db/migrations/4_add_org.rb
0atman: aha!

thomasiles:
Looks good!

No values for existing records - I don't think that matters though as we don't have any real users :)

0atman:
@aliuk2012 all done plz re-review!

aliuk2012:
@0atman I think something must of happened with the rebase because I see some of my commits in here that shouldn't relate to the changes you made. eg adding code coverage commits

0atman:
@aliuk2012 yeah, they're on main right?

aliuk2012:
🤦 Sorry, I just merged my PR that had migrations that ended up causing your branch to have conflicts....more reason to move to timestamp filename model

0atman:
@aliuk2012 ok this looks weird but can merge cleanly I think. Sorry about the weirdness

db/migrations/4_add_org.rb
aliuk2012: hmmm still seems to have pry

0atman:
@aliuk2012 removed pry for real

PR #47, opened 2022-08-03
Change create/delete page to alter order

thomasiles:
Pages form a linked list. Currently the client is responsible for
ensuring the consistency of the list.

This commit changes this by making sure new pages are 'appended' to the end
of the list. Deleting a page in the middle of the page list now connects
the two lists together. These changes mean there should only be one
connected list of pages.

This means clients are no longer able to create pages which random
'next' values. The update route hasn't been changed to keep that
possibility available, but maybe should be closed off too to ensure
consistency.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/lib/repositories/pages_repository.rb
+++ b/lib/repositories/pages_repository.rb
@@ -52,7 +52,7 @@ class Repositories::PagesRepository
   end
 
   def get_pages_in_form(form_id)
-    @database[:pages].where(form_id:).order(:next).all.map { |p| page_from_data(p) }
+    @database[:pages].where(form_id:).order(:id).all.map { |p| page_from_data(p) }
   end
 
   private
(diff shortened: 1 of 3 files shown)


lib/repositories/pages_repository.rb
```diff
@@ -31,11 +40,19 @@ def update(page)
   end
 
   def delete(page_id)
-    @database[:pages].where(id: page_id).delete
+    @database.transaction(isolation: :serializable) do
+      pages = @database[:pages]
+
+      # get the next value of our page and update any pages which pointed to us
+      # to that instead
```
DavidBiddle, line 47: Think this also needs to account for the case where the first page in a form is deleted - so the value that needs updating is the parent form's `start_page`

DavidBiddle: Looks like you're right about `start_page` just returning the lowest id so I'll resolve this and approve. At some point (probably when we add page reordering?) we'll need to change it to explicitly update `start_page` when the first page is created / deleted but don't need to worry about that yet

DavidBiddle (approved):
Looks good 🎉

PR #54, opened 2022-08-17
Fix adding next page to other forms

aliuk2012:
Relates to https://trello.com/c/30BWYuYQ/42-bug-next-page-value-not-being-set-correctly
When new pages were being created for a
specific form we were setting the next page
attribute for all pages even if the page
wasn't linked to the form that we were updating.
#### Checklist

    - [x] README.md
    - [x] Elsewhere (please link)


0atman (approved):
nice, looks good

PR #56, opened 2022-08-17
Add missing environment variable to manifest

DavidBiddle:
Fixes an issue with the previous sentry PR - errors in the dev enviroment were coming through as 'local', because the `PAAS_ENVIRONMENT` variable wasn't available in the app.

#### Checklist

- [n/a] I've linked this PR to the relevant issue (if mission work)
- [n/a] I've written unit tests for these changes (if code change)
    - [n/a] README.md

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,12 +1,13 @@
 ---
 applications:
-- name: forms-api-((PAAS_ENVIRONMENT))
-  memory: 256M
-  command: bundle exec rackup -o 0.0.0.0 -p $PORT
-  services:
-    - forms-api-((PAAS_ENVIRONMENT))-db
-    - forms-api-((PAAS_ENVIRONMENT))-splunk
-  env:
-    RACK_ENV: production
-    SENTRY_DSN: ((SENTRY_DSN))
-    API_KEY: ((API_KEY))
+  - name: forms-api-((PAAS_ENVIRONMENT))
+    memory: 256M
+    command: bundle exec rackup -o 0.0.0.0 -p $PORT
+    services:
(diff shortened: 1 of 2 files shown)


manifest.yml
```diff
@@ -1,12 +1,13 @@
...
+    memory: 256M
+    command: bundle exec rackup -o 0.0.0.0 -p $PORT
+    services:
+      - forms-api-((PAAS_ENVIRONMENT))-db
+      - forms-api-((PAAS_ENVIRONMENT))-splunk
+    env:
+      RACK_ENV: production
+      SENTRY_DSN: ((SENTRY_DSN))
+      API_KEY: ((API_KEY))
+      PAAS_ENVIRONMENT: ((PAAS_ENVIRONMENT))
```
0atman, line 13: is this indentation correct?

DavidBiddle: Not sure - VS Code did it automatically based on the editorconfig file. It's usually pretty good at not changing the  meaning when it does these things, and I don't _think_ Yaml cares about the difference between:
```
a
- b
```

and

```
a
  - b
```

but I'm happy to change it back if it breaks the deploy

DavidBiddle: I've just merged something similar in the admin and it looks like the deploy ran OK - https://github.com/alphagov/forms-runner/actions/runs/2876513134

DavidBiddle: Going to resolve this for now, we can revisit if we spot any issues (or if other people's editors start changing it back)

aliuk2012 (approved):

PR #58, opened 2022-08-18
Update sentry-ruby gem version

aliuk2012:
Sentry UI was complaining that we were using outdated SDK
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -125,11 +125,8 @@ GEM
       parser (>= 3.1.1.0)
     ruby-progressbar (1.11.0)
     ruby2_keywords (0.0.5)
-    sentry-ruby (5.3.0)
+    sentry-ruby (5.4.2)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -125,10 +125,10 @@ GEM
       parser (>= 3.1.1.0)
     ruby-progressbar (1.11.0)
     ruby2_keywords (0.0.5)
-    sentry-ruby (5.3.0)
+    sentry-ruby (5.3.1)
```
DavidBiddle, line 128: Could this not be 5.3.4 like the others? That would remove the `sentry-ruby-core` dependency too

aliuk2012: Good catch, odd that form-api pins gem versions but the other two apps don't

DavidBiddle (approved):

PR #63, opened 2022-08-23
Update next to next page

alice-carr:
We've decided to rename the `next` column in the pages table to `next_page`. This is because `next` is a reserved keyword in ruby and we want to save ourselves issues in the future. 

Before this is merged I'll need to make PRs in the admin and runner reflecting this change. 

I will also be updating this column to be an integer, not a string, but this will be part of a separate PR 

#### Checklist

### DIFF
--- a/db/migrations/9_rename_next_to_next_pages.rb
+++ b/db/migrations/9_rename_next_to_next_pages.rb
@@ -1,0 +1,9 @@
+Sequel.migration do
+  up do
+    rename_column :pages, :next, :next_page
+  end
+
+  down do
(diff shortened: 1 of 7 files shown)


0atman (approved):

aliuk2012 (approved):

aliuk2012:
I rebased this off main because github hasn't happy with the branch and said it was not in a mergeable state. I do have a copy of this branch before I rebase it should there be any issues but I did diff my local rebased version against the original branch on github and there was no difference to it.

PR #66, opened 2022-09-02
Rename migration files to help with sorting order

aliuk2012:
The problem with our existing migration file names its that
as soon as adding anything higher than 10 means that the sorting
of the migration files in the filesystem is jumbled up.

TBH, ideally, we would use DateTime stamped migration filenames to
avoid merge conflicts if other devs creating multiple migrations and
follows the pattern used by Rails.

Relates to https://trello.com/c/bQ5Jafsl/95-rename-api-migration-files-to-allow-use-to-create-999-migrations-without-issues-with-sorting

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


DavidBiddle (approved):
Looks good to me 🎉

DavidBiddle:
Just picking up on this: 
> TBH, ideally, we would use DateTime stamped migration filenames to
avoid merge conflicts if other devs creating multiple migrations and
follows the pattern used by Rails.
Agree on this, but might be even better if we could avoid writing our migrations by hand - surely there's a decent gem out there for this?

aliuk2012:
@0atman and I have been chatting about this issue....watch this space 😉

PR #67, opened 2022-09-05
Add form_slug to forms table

aliuk2012:
At the moment we rely on form id to find forms. We would like human-readable URLs. At the moment we are only concerned about the form name being used in the URL. 

There are obviously a lot of different ways this could have been approached but I decided to keep it simple and use [ActiveSupport parameterize method](https://api.rubyonrails.org/v7.0.3/classes/ActiveSupport/Inflector.html#method-i-parameterize) to tidy up the form name that the admin user has provided.

Related to https://trello.com/c/J5E6JuuY/92-use-human-readable-url-paths-for-forms-instead-of-id-based-paths
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


0atman (approved):
looks great, and I like the renumbering of migrations

PR #71, opened 2022-09-06
added api validation

0atman:
To allow a difference between form creators to have forms which are in a draft state/live state.

Add API validation on changing the live_at field to validate the submission_email, name, at least one page and privacy notice exist.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/lib/repositories/forms_repository.rb
+++ b/lib/repositories/forms_repository.rb
@@ -16,6 +16,14 @@ class Repositories::FormsRepository
   end
 
   def update(form)
+    if form[:live_at]
+      the_form = @database[:forms].where(id: form[:form_id])
+      raise "Invalid email" unless form[:submission_email].include?("@")
+      raise "Form has no name" unless form[:name]
+      raise "Form has no pages" unless @database[:pages].where(form_id: form[:form_id]).count > 0
+      raise "Form has a privacy policy" unless form[:privacy_policy_url]
+    end
+
     @database[:forms].where(id: form[:form_id]).update(
       name: form[:name],
       submission_email: form[:submission_email],
--- a/lib/repositories/forms_repository.rb
+++ b/lib/repositories/forms_repository.rb
@@ -18,10 +18,18 @@ class Repositories::FormsRepository
   def update(form)
     if form[:live_at]
       the_form = @database[:forms].where(id: form[:form_id])
-      raise "Invalid email" unless form[:submission_email].include?("@")
-      raise "Form has no name" unless form[:name]
-      raise "Form has no pages" unless @database[:pages].where(form_id: form[:form_id]).count > 0
-      raise "Form has a privacy policy" unless form[:privacy_policy_url]
+      if not form[:submission_email].include?("@")
+        error!({ error: 'invalid', detail: 'email field' }, 400)
+      end
+      if not form[:name]
+        error!({ error: 'missing', detail: 'form name field' }, 400)
+      end
+      if @database[:pages].where(form_id: form[:form_id]).count < 1
+        error!({ error: 'missing', detail: 'form has no pages' }, 400)
+      end
+      if not form[:privacy_policy_url]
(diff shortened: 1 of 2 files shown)


lib/repositories/forms_repository.rb
```diff
@@ -16,6 +16,14 @@ def get_by_org(org)
   end
 
   def update(form)
+    if form[:live_at]
+      the_form = @database[:forms].where(id: form[:form_id])
+      raise "Invalid email" unless form[:submission_email].include?("@")
+      raise "Form has no name" unless form[:name]
+      raise "Form has no pages" unless @database[:pages].where(form_id: form[:form_id]).count > 0
+      raise "Form has a privacy policy" unless form[:privacy_policy_url]
```
aliuk2012, line 24:
```suggestion
      raise "Form has no privacy policy URL" unless form[:privacy_policy_url]
```

0atman: Thanks so much Al, but maybe we'll have to throw (ha) away this exception system, which is what I've been noodling with? Lets talk about it

0atman:
OK NEW PLAN.

After talking with @aliuk2012 I'm going to use native grape exceptions, which the router will be able to correctly interpret into HTTP errors. :crossed_fingers: 
https://github.com/ruby-grape/grape#exception-handling

0atman:
@aliuk2012 sorry I ran out of time here, I added a failing test to show the problem, but my ruby is not as good enough to figure out how to import the missing `error!` function!

```
     NoMethodError:
       undefined method `error!' for #<Repositories::FormsRepository:0x00007ff887f8ae58 @database=#<Sequel::Postgres::Database: "postgres://postgres:password@localhost/postgres_test">>
     
               error!({ error: 'invalid', detail: 'email field' }, 400)
               ^^^^^^
     # ./lib/repositories/forms_repository.rb:22:in `update'
     # ./spec/repositories/forms_repository_spec.rb:92:in `block (3 levels) in <top (required)>'
     # ./spec/database_context.rb:5:in `block (3 levels) in <top (required)>'
     # ./spec/database_context.rb:5:in `block (2 levels) in <top (required)>'
```

aliuk2012:
Its a good thing we did attempt to write some unhappy path tests because the reason the tests are failing is because the `error!` method does not exist in the repository class. This method is actually part of Grape which we arent importing/inheriting from which is why it throws this error and would have done so when running the app normally.

0atman:
Thanks for picking this up @aliuk2012 
I think what is needed is instead of using the non-accessible `error!()` method, we throw the appropriate one of these exceptions:
https://github.com/ruby-grape/grape/tree/master/lib/grape/exceptions

which, if we're lucky will be handled by the grape router automatically, giving us the correct http error codes.

lib/repositories/forms_repository.rb
```diff
@@ -16,6 +16,22 @@ def get_by_org(org)
...
+      if not form[:name]
+        error!({ error: 'missing', detail: 'form name field' }, 400)
+      end
+      if @database[:pages].where(form_id: form[:form_id]).count < 1
+        error!({ error: 'missing', detail: 'form has no pages' }, 400)
+      end
+      if not form[:privacy_policy_url]
+        error!({ error: 'missing', detail: 'form has no privacy policy url' }, 400)
+      end
+    end
```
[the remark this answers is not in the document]
0atman: Turns out `error!()` isn't in scope.

I think something like a `ValidationError` would work https://github.com/ruby-grape/grape/blob/master/lib/grape/exceptions/validation_errors.rb

aliuk2012:
I'm closing this PR now because as a team we discussed and agreed the following:

As our api is not public and only forms-admin & forms-runner are using it we would add the validation in those applications so that requests are not even made to the forms-api if the users data is incorrect. Should we want to make the api public we would then consider creating different read (or write if required) endpoints for public use.

PR #84, opened 2022-09-30
Minor amendments to the Dockerfile

danworth:
- Add and use a non-root user named ruby
- Run the server to listen on 0.0.0.0 so it'll work without needing host networking

Using a non-root user is more secure. Currently the image is only used for Cypress testing forms-admin where the network mode is set to host. In order for this image to work without host networking the server needs to listen on `0.0.0.0` which will bind to all network interfaces in the network namespace.

### Testing
This builds ok on my M1 mac. We'll add github actions to run the build of these Dockerfiles on any change.

There's a small risk that this might break the forms-admin cypress tests but since the exposed port remains as `9292` it should be ok.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,20 @@
-FROM ruby:3.1.1-alpine
+# NOTE: Before using in production we must consider the need to pin
+# images to a SHA
+FROM ruby:3.1.1-alpine3.15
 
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
builds ok on my machine 👍

SamJamCul (approved):
Builds on my M1 👍🏻

PR #91, opened 2022-10-07
Switch is_optional from TrueClass to String

aliuk2012:
None of our other params have specific types. There also had a problem with the requests coming from forms-admin where the value was being sent as strings in the call. The database column is setup as boolean so should fail any requests that try to set it to non-boolean values.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


DavidBiddle (approved):
Looks good to me (even if it does seem strange that Ruby doesn't provide a nice boiolean class we can use here!) 🎉

PR #108, opened 2022-11-08
Revert "Bump grape-swagger from 1.4.2 to 1.5.0"

aliuk2012:
Reverts alphagov/forms-api#81

Still seems to confuse PaaS with server to use. Will have another look when I got more time 

https://github.com/alphagov/forms-deploy/actions/runs/3418780768/jobs/5691696526 failing action


thomasiles (approved):
👍

PR #116, opened 2022-11-21
Add feature: reordering the pages in a form

thomasiles:
Currently users cannot change the order of pages in a form - they must
delete and recreate them.

The API returns pages ordered by id, when serving a form. This order is
used to display the pages in the admin's page list. This works because
the page.id always increases, and because the user has no way of moving
pages.

Pages in a form use the .next_page attribute to form an implicit linked
list. The start_page acts as the head of the list.

The runner uses the `start_page` attribute, which is not stored in the
database and is calculated by the API when retrieving a form, based on
lowest page id. It then moves from page to page using the id found in
the next_page attribute.

Adding reordering
This PR introduces two new endpoints for moving a page earlier and later
in the order of a form. This can then then be used to move pages up and
down in the admin.

To do this, we need to do three things:
- Change how the order of pages is calculated
- Change how the start_page is calculated
- Add a new endpoint for moving a page in the list

This commit introduces a new column on the forms table, page_order,
which contains a pg_array of page_ids, in the order they appear in the
form. This is used for all page ordering operations, including
calculating the .next_page attribute of pages. The next_page column
could be dropped. This commit doesn't do that in case we need to roll
back.

The .next_page attribute of pages used to be stored in column and was
altered on page creation and deletion. The new scheme calculates the
next_page attribute when fetch a single or multiple forms. Calculating
this doesn't seem to be too expensive but given how often all the pages
of a form are fetched, we should monitor it. If it does become an
issue, we could write the next_page attributes when order has been
modified. As we are considering other changes to the data model and API
I think it's worth holding of unless we know it's an issue.

Two new routes have been added to the API for moving the order of pages
by one place.

```
PUT /api/v1/forms/<form_id>/pages/<page_id>/up
PUT /api/v1/forms/<form_id>/pages/<page_id>/down
```

Whenever API routes are called which might result in changes to the order, the
page_order is modified too.

There are no constraints on the page_order column which prevent the data
being inconsistent. Care has to be taken when changing it to stop it
containing duplicates, referencing pages which don't belong to the form
or pages which don't exist. page_order is not exposed by the API for
this reason. If more feaures are added to modify the array, we should
add more validation checks when it's changed.

Effect on the runner
-----------
The runner is quite robust if a creator changes the order of a form
whilst a user is filling out a form. The worst consequence seems to what
appears to be repeated questions from the point of view of the filler
inner. These questions will already be filled out and the filler can
continue through the form.

Performance
-----------

I ran some very unscientific tests with forms with a very large number
of pages. The results are below. There was nothing alarming in them -
but I don't think they tell us very much either.

```bash

DEFAULTVALUE='government-digital-service'
curl --include --silent --show-error \
  --request POST \
  -H 'Content-Type: application/json' \
  -H "X-Api-Token: ${API_KEY}" \
  --url "$API_BASE/forms/$1/pages" -d @-
```

```bash
printf '{"question_text": "q %s", "answer_type": "single_line"}' $i | ./add_page.sh 1
```

```bash

DEFAULTVALUE='government-digital-service'
curl -vv last_trace.curl \
  --include --silent --show-error \
  --request GET \
  -H 'Content-Type: application/json' \
  -H "X-Api-Token: ${API_KEY}" \
  --url "$API_BASE/forms/$1/pages" | tee last_response.http | sed '1,/^\r\{0,1\}$/d' | jq
```

Number of forms, Seconds to perfom a request to fetch all pages of the form
1000  0.161
2000  0.281
3000  0.369
4000  0.409
5000  0.528
6000  0.598
7000  0.695
8000  0.958
9000  0.958
10000 1.231

Trello card: https://trello.com/c/vUh413Cv/296-spike-exploration-data-model-redesign-reorder-questions

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/db/migrations/019_add_page_order_to_form.rb
+++ b/db/migrations/019_add_page_order_to_form.rb
@@ -1,0 +1,21 @@
+Sequel.migration do
+  up do
+    add_column :forms, :page_order, "integer[]", default: []
+
+    # Before this change, the order of pages is defined by page id, which
+    # increases with every new page. Running this update collects the page ids for each
+    # form and adds them to the form as an array, in the correct order.
+    update_existing = <<~SQL
+      UPDATE forms
+         SET page_order=pages_id_order
+        FROM (SELECT form_id, array_agg(id ORDER BY id) pages_id_order FROM pages GROUP BY form_id) forms_pages
+       WHERE forms_pages.form_id=forms.id
+    SQL
+
+    run update_existing
+  end
+
+  down do
+    drop_column :forms, :page_order
+  end
+end
--- a/lib/api_v1.rb
+++ b/lib/api_v1.rb
@@ -62,8 +62,9 @@ class APIv1 < Grape::API
         page_repository = Repositories::PagesRepository.new(@database)
 
         form = repository.get(params[:form_id])
-        pages = page_repository.get_pages_in_form(params[:form_id]).sort_by(&:id)
-        form[:start_page] = (pages.first.id if pages.any?)
+
+        form[:start_page] = page_repository.get_pages_in_form(params[:form_id])&.first&.id
+
         form
       end
 
@@ -106,7 +107,7 @@ class APIv1 < Grape::API
         desc "Return all pages for the form"
         get do
           repository = Repositories::PagesRepository.new(@database)
-          repository.get_pages_in_form(params[:form_id]).sort_by(&:id).map(&:to_h)
+          repository.get_pages_in_form(params[:form_id]).map(&:to_h)
         end
 
         desc "Create a new page."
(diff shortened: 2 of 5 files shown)


danworth:
>to save the page order in a jsonb array in form

Yes, I wonder if saving the page order as a separate attribute on the form is the best solution. It would mean we don't need to update every page whenever the order is changed (as we do with the current linked-list approach). I'm not familiar enough with the runner or api structure to know the full impact of this but as a change to the underlying data model it seems to offer the most flexibility and could possibly be extended to handle conditionally optional questions.

DavidBiddle (approved):
This looks good to me!

I guess one thing to watch out for is how this will work with our eventual skip logic implementation, but it's hard to make any concrete predictions on that now.

0atman (approved):
makes sense, clear migration tests :ok_hand:

PR #117, opened 2022-11-25
Add selection answer type and its options

DavidBiddle:
Adds an `answer_settings` JSON column to the pages object. When returned from the API it will look something like this:

```json
"answer_settings": {
  "allow_multiple_answers": true,
  "selection_options": [
    {"name": "Option 1"},
    {"name": "Option 2"},
  ]
}
```

The intent is for this field to be flexible for different answer types - for example, this field will likely end up containing flags for autofill related settings for other fields.  

Trello card: https://trello.com/c/bSqKTDu8/224-add-selection-question-type-to-api

#### Checklist

  - [n/a] README.md

### DIFF
--- a/db/migrations/018_add_answer_options_to_page.rb
+++ b/db/migrations/018_add_answer_options_to_page.rb
@@ -1,0 +1,8 @@
+Sequel.migration do
+  up do
+    add_column :pages, :answer_settings, :jsonb
+  end
+  down do
+    drop_column :pages, :answer_settings
+  end
+end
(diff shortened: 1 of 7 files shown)


thomasiles (approved):
looks good! I guess having the answer_settings as a nested object is easy to work with in the admin and runner?

DavidBiddle:
@thomasiles It's reasonably easy I think, ActiveResource does a good job reading it and mapping it to a class, setting up the fields in a form is slightly fiddlier.

PR #118, opened 2022-11-30
set rackup bind host to 0.0.0.0

0atman:
A simple tweak so that the dockerised admin and runner can connect to our local API


aliuk2012:
Closing this as we no longer using rack but rails server now

PR #125, opened 2022-12-11
Switch from ruby-grape to rails-api

aliuk2012:
We plan to move away from ruby-grape to rails-api purely because most of
the team aren't familiar with ruby and its making it hard for all devs
to contribute to forms service when the service has a combination of
two very different frameworks and conventions.

We agreed that we would continue to use 3 seperate microservices
(forms-admin, forms-api & forms-runner) but this will hopefully reduce
the need for us to make it monolith. We will monitor the situation and
if we do still need to merge all the microservices we would be
in a better position as all the services would be using rails conventions#### What problem does the pull request solve?

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


config/puma.rb
```diff
@@ -0,0 +1,43 @@
...
+threads min_threads_count, max_threads_count
+
+# Specifies the `worker_timeout` threshold that Puma will use to wait before
+# terminating a worker in development environments.
+#
+worker_timeout 3600 if ENV.fetch("RAILS_ENV", "development") == "development"
+
+# Specifies the `port` that Puma will listen on to receive requests; default is 3000.
+#
+port ENV.fetch("PORT") { 9292 }
```
aliuk2012, line 18: still uses port 9292

config/database.yml
```diff
@@ -0,0 +1,86 @@
...
+default: &default
+  adapter: postgresql
+  encoding: unicode
+  # For details on connection pooling, see Rails configuration guide
+  # https://guides.rubyonrails.org/configuring.html#database-pooling
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+
+development:
+  <<: *default
+  database: postgres
```
aliuk2012, line 26: set it to this DB so we can test locally using existing db

.env.template
```diff
@@ -1,3 +1,2 @@
-DATABASE_URL=DATABASE_URL
```
aliuk2012, line 1: we dont really need to set this locally unless we using a non-standard db configuration....i.e changed default port 5432 etc.

PR #161, opened 2023-01-03
Add dockerfile

thomasiles:
This commit add rails compatible Dockerfile.

It's based on the one used in the admin project, with the node and asset setup removed.

See https://github.com/alphagov/forms-deploy/pull/64 for use in the local environment.

Trello card: https://trello.com/c/MhFAtX8B/413-add-docker-to-rails-api-branch

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


.github/workflows/docker_build.yml
```diff
@@ -1,9 +1,9 @@
 name: "Docker build"
 on:
   push:
-    branches: [main]
+    branches: [ rails-api ]
```
aliuk2012, line 4:
```suggestion
    branches: [ main, rails-api ]
```
I think we should include both for the time being. It shouldn't run on main branch just yet because the actions only really run from whatever is declared in main branch

aliuk2012 (approved):

PR #165, opened 2023-01-05
Reset pages complete

thomasiles:
#  Update forms question_section_completed
When a form creator makes changes to a page, the question_section_completed value for a form should be reset to false. 
This is used to show the status in the tasklist.

This commit updates the value when pages are created, updated or deleted.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/api/v1/pages_controller.rb
+++ b/app/controllers/api/v1/pages_controller.rb
@@ -10,6 +10,7 @@ class Api::V1::PagesController < ApplicationController
     page = form.pages.new(page_params)
 
     if page.save
+      form.update!(question_section_completed: false)
       render json: { id: page.id }, status: :created
     else
       render json: page.errors.to_json, status: :bad_request
@@ -23,6 +24,7 @@ class Api::V1::PagesController < ApplicationController
 
   def update
     if page.update(page_params)
+      form.update!(question_section_completed: false)
       render json: { success: true }.to_json, status: :ok
     else
       render json: page.errors.to_json, status: :bad_request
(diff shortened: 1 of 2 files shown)


spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -136,14 +144,17 @@
...
+      expect(page1.reload.question_text).to eq("updated page title")
+      expect(form.reload.question_section_completed).to be false
     end
 
     it "fields not in the params are cleared" do
       expect(page1.hint_text).to be_nil
     end
 
+    it "sets question_section_completed to false" do
+    end
```
aliuk2012, line 156: ❓

app/controllers/api/v1/pages_controller.rb
```diff
@@ -22,7 +23,10 @@ def show
   end
 
   def update
-    if page.update(page_params)
+    page.update!(page_params)
+
+    if page.valid?
```
aliuk2012, line 28: could we maybe using `.save` instead of `.update` like we do for create action? 

`.save` creates the record if it doesn't exist or it updates it.

thomasiles: I guess so - do we lose anything by doing that though? If an API user wants to create a page, they can use create or update. If we want to add logic in the future which only happens on create would that make it harder? Would logging be more difficult if update can also create pages?

aliuk2012: I dont think it would make it harder to log events because the user would still be calling the various create/update controller action

PR #174, opened 2023-01-11
WIP: Add paper trail

thomasiles:
# Add basic papertrail support for Form and Pages model

Adds [the papertail gem](https://github.com/paper-trail-gem/paper_trail) and enables version tracking on the Form & Page model.

```
bundle exec rails generate paper_trail:install --with-changes
```

The `object` and `object_changes` columns have [been set to jsonb](https://github.com/paper-trail-gem/paper_trail#6b-custom-serializer).

Trello card: https://trello.com/c/7T5Yp818/426-add-paper-trail-gem-to-audit-and-changes-to-forms-table

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,6 +17,10 @@ Style/NumericLiterals:
   Exclude:
     - spec/migrations/**/*
 
+Rails/CreateTableWithTimestamps:
+  Exclude:
+    - db/migrate/20230112140812_create_version_associations.rb
+
 # **************************************************************
 # TRY NOT TO ADD OVERRIDES IN THIS FILE
 #
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -7,6 +7,8 @@ abort("The Rails environment is running in production mode!") if Rails.env.produ
 require "rspec/rails"
 # Add additional requires below this line. Rails is not loaded until this point!
 
+require "paper_trail/frameworks/rspec"
+
 # Requires supporting ruby files with custom matchers and macros, etc, in
 # spec/support/ and its subdirectories. Files matching `spec/**/*_spec.rb` are
 # run as spec files by default. This means that files in spec/support that end
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem "config"
 
 # Auditing model changes and logging versions
 gem "paper_trail"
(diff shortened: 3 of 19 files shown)


thomasiles:
A couple of things to think about:
- How does deleting an object work? Should we switch to soft deletion or is it not something we need to worry about? Would it make doing later version features easier or just make everything more complicated?
- Should we use a [custom class](https://github.com/paper-trail-gem/paper_trail#5b-configuring-the-versions-association) for the version - it sounds like it would be handy 
- [Optimistic locking](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html). Can we add it to the API? The publishing API might be a [good reference](https://github.com/alphagov/publishing-api/blob/main/docs/api.md#optimistic-locking-previous_version)
- Adding whodunnit info to versions. Can we add user uuid to a header in the admin and then pick it up in the API as a way to collect which users made the change? This is easy to implement but there may be downsides
- Does it make sense to deploy this before we have a solution for storing page information?

.rubocop.yml
```diff
@@ -17,6 +17,10 @@ Style/NumericLiterals:
   Exclude:
     - spec/migrations/**/*
 
+Rails/CreateTableWithTimestamps:
+  Exclude:
+    - db/migrate/20230112140812_create_version_associations.rb
```
aliuk2012, line 22: this is a migration that was created by a papertrail associations gem.

spec/rails_helper.rb
```diff
@@ -20,7 +22,7 @@
 # directory. Alternatively, in the individual `*_spec.rb` files, manually
 # require only the support files necessary.
 #
-# Dir[Rails.root.join('spec', 'support', '**', '*.rb')].sort.each { |f| require f }
+# Dir[Rails.root.join("spec', 'support', '**', '*.rb')].sort.each { |f| require f }
```
lfdebrux, line 25:
```suggestion
# Dir[Rails.root.join('spec', 'support', '**', '*.rb')].sort.each { |f| require f }
```

aliuk2012:
We decided to abandon the idea of using papertrail for draft/live because of the complexity that comes with associations. Instead we created our own implementation that copys forms & pages objects adn stores it as json in it own table.

https://github.com/alphagov/forms-api/pull/207

PR #181, opened 2023-01-20
Strong params changes for name input

DavidBiddle:
Adds the new name type to the list of allowed answer types, and allows it to send an `answer_settings` hash in the strong params.

Trello card: https://trello.com/c/efuBewje/384-persons-name-implementation-of-autofill-work-in-production

#### Checklist

- [n/a] I've written unit tests for these changes (if code change)
  - [n/a] README.md

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -2,7 +2,7 @@ class Page < ApplicationRecord
   belongs_to :form
   acts_as_list scope: :form
 
-  ANSWER_TYPES = %w[single_line number address date email national_insurance_number phone_number long_text selection organisation_name text].freeze
+  ANSWER_TYPES = %w[single_line number address date email national_insurance_number phone_number long_text selection organisation_name text name].freeze
 
   validates :question_text, presence: true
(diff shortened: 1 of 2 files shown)


app/models/page.rb
```diff
@@ -2,7 +2,7 @@ class Page < ApplicationRecord
   belongs_to :form
   acts_as_list scope: :form
 
-  ANSWER_TYPES = %w[single_line number address date email national_insurance_number phone_number long_text selection organisation_name text].freeze
+  ANSWER_TYPES = %w[single_line number address date email national_insurance_number phone_number long_text selection organisation_name text name].freeze
```
aliuk2012, line 5: Shouldn't there be some tests around these new types?

DavidBiddle: @aliuk2012 Think this is already covered - this `ANSWER_TYPES` constant is sampled in the page factory so the new answer types will be tested in the page model tests

SamJamCul (approved):

PR #184, opened 2023-01-26
Add migrations for autofill answer types

DavidBiddle:

Adds migrations for the pre-existing answer types (address, date, single line of text, multiple lines of text) that have had answer settings pages added.

Also includes some logic to allow the API to handle pages with the legacy `single_line` and `long_text` inputs in the `create`, `update` and `show` actions. This is behind a feature flag `accept_legacy_answer_types` - this will default to `true` and should be changed to `false` as soon as the `autofill_answer_types` feature flag in admin is set to `true`. We only need to add this for the text answer types - `address` and `date` are already sending `answer_settings` to the API. 

Still to do:

- [x] Consider whether to merge these into a single migration, rather than three separate ones
- [x] Add tests for the migration

Trello card: https://trello.com/c/RDzeLJqz/461-write-migrations-for-changed-autofill-fields

#### Checklist

  - [n/a] README.md

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,9 @@ inherit_mode:
   merge:
     - Exclude
 
+Style/NumericLiterals:
+  Exclude:
+    - spec/migrations/**/*
 
 # **************************************************************
 # TRY NOT TO ADD OVERRIDES IN THIS FILE
--- a/app/controllers/api/v1/pages_controller.rb
+++ b/app/controllers/api/v1/pages_controller.rb
@@ -123,19 +123,12 @@ private
     when "long_text"
       page.answer_settings = { input_type: "long_text" }
       page.answer_type = "text"
-    when "address"
-      page.answer_settings ||= { input_type: { uk_address: "true", international_address: "false" } }
-    when "date"
-      page.answer_settings ||= { input_type: "other_date" }
     end
   end
 
   def display_new_answer_types_in_old_format(page)
-    case page.answer_type
-    when "text"
+    if page.answer_type == "text"
       page.answer_type = page.answer_settings["input_type"]
-      page.answer_settings = nil
-    when "date", "address"
       page.answer_settings = nil
     end
   end
--- a/app/controllers/api/v1/pages_controller.rb
+++ b/app/controllers/api/v1/pages_controller.rb
@@ -115,21 +115,21 @@ private
     render json: { error: exception.message }, status: :bad_request
   end
 
-  def convert_old_answer_types_to_new_format(page)
-    case page.answer_type
+  def convert_old_answer_types_to_new_format(page_object)
+    case page_object.answer_type
     when "single_line"
-      page.answer_settings = { input_type: "single_line" }
-      page.answer_type = "text"
+      page_object.answer_settings = { input_type: "single_line" }
+      page_object.answer_type = "text"
     when "long_text"
-      page.answer_settings = { input_type: "long_text" }
-      page.answer_type = "text"
+      page_object.answer_settings = { input_type: "long_text" }
+      page_object.answer_type = "text"
     end
   end
 
-  def display_new_answer_types_in_old_format(page)
-    if page.answer_type == "text"
-      page.answer_type = page.answer_settings["input_type"]
-      page.answer_settings = nil
+  def display_new_answer_types_in_old_format(page_object)
+    if page_object.answer_type == "text"
+      page_object.answer_type = page_object.answer_settings["input_type"]
+      page_object.answer_settings = nil
     end
   end
 end
--- a/app/controllers/api/v1/pages_controller.rb
+++ b/app/controllers/api/v1/pages_controller.rb
@@ -9,6 +9,10 @@ class Api::V1::PagesController < ApplicationController
   def create
     new_page = form.pages.new(page_params)
 
+    if FeatureService.enabled?(:accept_legacy_answer_types)
+      convert_old_answer_types_to_new_format(new_page)
+    end
+
     if new_page.save
       form.update!(question_section_completed: false)
       render json: { id: new_page.id }, status: :created
@@ -18,11 +22,21 @@ class Api::V1::PagesController < ApplicationController
   end
 
   def show
+    if FeatureService.enabled?(:accept_legacy_answer_types)
+      display_new_answer_types_in_old_format(page)
+    end
+
     render json: page.to_json, status: :ok
   end
 
   def update
-    if page.update(page_params)
+    page.assign_attributes(page_params)
+
+    if FeatureService.enabled?(:accept_legacy_answer_types)
+      convert_old_answer_types_to_new_format(page)
+    end
+
+    if page.save
       form.update!(question_section_completed: false)
       render json: { success: true }.to_json, status: :ok
     else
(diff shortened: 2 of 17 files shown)


aliuk2012:
Once this PR is merged does it mean we can get rid of these sort of checks in the codebase?

`answer_settings.present? && answer_settings&.`

app/controllers/api/v1/pages_controller.rb
```diff
@@ -9,6 +9,10 @@ def index
   def create
     new_page = form.pages.new(page_params)
 
+    if FeatureService.enabled?(:accept_legacy_answer_types)
+      convert_old_answer_types_to_new_format(new_page)
+    end
```
aliuk2012, line 14: Does this work? Looking at the method, it seems to convert `page` rather than `new_page`.

app/controllers/api/v1/pages_controller.rb
```diff
@@ -100,4 +114,22 @@ def not_found
   def missing_parameter(exception)
     render json: { error: exception.message }, status: :bad_request
   end
+
+  def convert_old_answer_types_to_new_format(page)
```
aliuk2012, line 118: could we rename `page` as we already have a private method in this controller thats called `page` which makes its confusing to follow and understand that we are modifying the object thats passed to this method rather the `@page` which the page method would have set.

DavidBiddle:
@aliuk2012 YEs, think we should be able to get rid of the `answer_settings` check at least and probably the `input_type` one as well.

app/controllers/api/v1/pages_controller.rb
DavidBiddle: Done - called it `page_object` to make clear that's what it's expecting

aliuk2012:
Could we split out the feature flag commits into a separate PR? It looks good to go and I could really do with it for the draft/live feature.

spec/migrations/convert_address_to_uk_address_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:previous_version) { 20230126133557 }
+
+  describe "#up" do
+    before do
+      ActiveRecord::Migration.suppress_messages do
+        ActiveRecord::Migrator.new(:down, migrations, schema_migration, previous_version).migrate
+      end
+    end
+
+    it "converts existing date input types to ‘other_date‘ when nil" do
```
DavidBiddle, line 18:
```suggestion
    it "converts existing address input types to uk addresses when nil" do
```

.rubocop.yml
```diff
@@ -13,6 +13,11 @@ inherit_mode:
   merge:
     - Exclude
 
+Style/NumericLiterals:
+  Exclude:
+    - spec/migrations/convert_address_to_uk_address_spec.rb
+    - spec/migrations/convert_date_to_other_date_spec.rb
+    - spec/migrations/convert_old_text_fields_to_new_text_fields_spec.rb
```
aliuk2012, line 20: Might be worth excluding anything in the `spec/migrations` folder something like `spec/migrations/**/*' should do it rather htan manually ignoring every file that will use a version id

db/migrate/20230126133557_convert_date_to_other_date.rb
```diff
@@ -0,0 +1,9 @@
+class ConvertDateToOtherDate < ActiveRecord::Migration[7.0]
+  def up
+    Page.where(answer_type: "date", answer_settings: nil).update_all(answer_settings: { input_type: "other_date" })
+  end
+
+  def down
+    Page.where(answer_type: "date", answer_settings: { input_type: "other_date" }).update_all(answer_settings: nil)
```
danworth, line 7: Is there any chance that there would have already been pages with answer type 'date' that already had input_type of "other_date" set via forms-admin before the `up` migration was run, and thus would be incorrectly updated by this rollback `down` migration? Might not be important just curious if the `down` will only ever reverse the `up` or if could update additional entries not originally modified by the `up`? Even if that is the case, I'm not sure how important that would be but thought worth asking :)

config/settings.yml
```diff
@@ -9,6 +9,6 @@ sentry:
   dsn:
   environment: local
 
-# The feature service needs at least one feature in order to establish the pattern for features, this can be replaced with a real feature when we have one
 features:
-  example_feature: false
+  # accept_legacy_answer_types should be set to false when the autocomplete_answer_types flag in admin is set to true
+  accept_legacy_answer_types: true
```
aliuk2012, line 14: Could we add a spec for this?

db/migrate/20230126133557_convert_date_to_other_date.rb
DavidBiddle, line 7: I'm in 2 minds as to whether we should also perform the `down` migration on dates of birth - we'd be losing data  about what kind of date it is, but I'm not sure we ever want to be in a state where some dates are in the old format and some are in the new one:
```suggestion
    Page.where(answer_type: "date", answer_settings: { input_type: "other_date" }).update_all(answer_settings: nil)
    Page.where(answer_type: "date", answer_settings: { input_type: "date_of_birth" }).update_all(answer_settings: nil)
```

config/settings.yml
aliuk2012: example https://github.com/alphagov/forms-admin/blob/main/spec/config/settings_spec.rb#L21-L27

It helps ensure that someone doesn't just accidentally removes or disables a feature flag without thinking about it

db/migrate/20230126135721_convert_address_to_uk_address.rb
```diff
@@ -0,0 +1,9 @@
+class ConvertAddressToUkAddress < ActiveRecord::Migration[7.0]
+  def up
+    Page.where(answer_type: "address", answer_settings: nil).update_all(answer_settings: { input_type: { uk_address: "true", international_address: "false" } })
+  end
+
+  def down
+    Page.where(answer_type: "address", answer_settings: { input_type: { uk_address: "true", international_address: "false" } }).update_all(answer_settings: nil)
```
DavidBiddle, line 7: As with the dates - maybe this should cover all of our cases:
```suggestion
    Page.where(answer_type: "address", answer_settings: { input_type: { uk_address: "true", international_address: "false" } }).update_all(answer_settings: nil)
    Page.where(answer_type: "address", answer_settings: { input_type: { uk_address: "false", international_address: "true" } }).update_all(answer_settings: nil)
    Page.where(answer_type: "address", answer_settings: { input_type: { uk_address: "true", international_address: "true" } }).update_all(answer_settings: nil)
```

db/migrate/20230126133557_convert_date_to_other_date.rb
aliuk2012: Good point, I agree, we wouldn't want to be in a mix of formats (i.e find all date types and revert to "old format")

```suggestion
    Page.where(answer_type: "date").update_all(answer_settings: nil)
```

db/migrate/20230126135721_convert_address_to_uk_address.rb
aliuk2012: Same as dates
```suggestion
    Page.where(answer_type: "address").update_all(answer_settings: nil)
```

db/migrate/20230126133557_convert_date_to_other_date.rb
SamJamCul: This is something we discussed - we kind of concluded that there isn't really a way to revert certain migrations without storing historical data for them. This either means we explicitly build them without any reverse gear, or we try and get them to revert as much as possible so that anything that could have been a result of the is moved back. 

I don't think in this case it matters, because flipping this migration back and forth will simply bring everything in line - everything that's up to date already will be unchanged, and moving it back and forth will still retain the same detail.

spec/migrations/convert_address_to_uk_address_spec.rb
SamJamCul: sorted, good old copy paste

.rubocop.yml
SamJamCul: Done, makes sense to me!

config/settings.yml
SamJamCul: sure thing!

aliuk2012 (approved):

PR #187, opened 2023-01-31
Feature service

SamJamCul:
This PR adds a feature service to the API - it was part of this PR https://github.com/alphagov/forms-api/pull/184

The original work requires the feature flag service, but this is now going to be introduced in this new PR, which will be merged in prior so that other work that relies on having a feature service can take place.


aliuk2012 (approved):
Added some documentation to README but otherwise good to go.

PR #202, opened 2023-02-15
Rename development database to be more consistent

lfdebrux:
Suffix `_development` to the name of the development database to match the pattern used on other databases in this app and the forms-admin app.

The name of the development database is currently a little surprising.

#### Checklist

- [ ] ~I've linked this PR to the relevant issue (if mission work)~
- [ ] ~I've written unit tests for these changes (if code change)~
  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -23,7 +23,7 @@ default: &default
 
 development:
   <<: *default
-  database: forms_api
+  database: forms_api_development
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.


config/database.yml
```diff
@@ -23,7 +23,7 @@ default: &default
 
 development:
   <<: *default
-  database: forms_api
+  database: forms_api_development
```
aliuk2012, line 26: ❤️ , however we might have to change some dockerfile config to ensure this doesn't break docker builds

lfdebrux: Ah, good catch. I looked at forms_deploy/local just now, it looks like the configuration is already slightly different from what is in the Rails config files:

```
forms-deploy on  main at 09:52:13
% git grep forms_admin local
local/docker-compose.yml:      - DATABASE_URL=postgres://forms_admin:forms_admin@postgres/forms_admin
local/postgres/initdb.d/init.sql:CREATE USER forms_admin WITH PASSWORD 'forms_admin' CREATEDB;
local/postgres/initdb.d/init.sql:CREATE DATABASE forms_admin;
local/postgres/initdb.d/init.sql:GRANT ALL PRIVILEGES ON DATABASE forms_admin TO forms_admin;
local/postgres/initdb.d/init.sql:ALTER DATABASE forms_admin OWNER TO forms_admin;
```

So I don't think this PR will necessarily make things any worse, but we should probably make the change in forms_deploy to make things even more consistent.

lfdebrux:
@danworth is this going to affect the AWS stuff?

danworth:
> @danworth is this going to affect the AWS stuff?

Thanks for checking Laurence. We override the DB config by setting the `DATABASE_URL` env var which takes precedence over the config files. Also we set `RACK_ENV` to `production` in all AWS and PaaS deployments so AFAIK it wouldn't be using the `development` config settings anyway.
https://github.com/alphagov/forms-deploy/blob/ffaf19399a01afff1cb99343a2533f91af7744c8/infra/modules/forms-api/main.tf#L23.

aliuk2012 (approved):
Looks good to me. Might be worth mentioning it to other devs in slack. If anyone is using default rails config then will need to run  `rail db:drop && rails db:create && rails db:migrate` because the app will be pointing to a new different DB.

PR #207, opened 2023-02-20
Creating made_live_forms

aliuk2012:

Creates a new published form table to store a version of forms and pages as a single JSON blob.

Tweaks how make_live and live endpoints work to use the stored live version JSON blobs.

Trello card: https://trello.com/c/7T5Yp818/426-implement-draft-live-versioning-in-backend

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -2,7 +2,7 @@ class Form < ApplicationRecord
   has_paper_trail
 
   has_many :pages, -> { order(position: :asc) }, dependent: :destroy
-  has_many :made_live_forms
+  has_many :made_live_forms, dependent: :restrict_with_exception
 
   validates :org, :name, presence: true
   def start_page
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -11,10 +11,14 @@ class Form < ApplicationRecord
 
   def make_live!
     update!(live_at: Time.zone.now)
+
+    made_live_forms.create!(json_form_blob: to_json(include: [:pages])) if FeatureService.enabled?(:draft_live_versioning)
   end
 
   def live_version
-    snapshot
+    return snapshot unless FeatureService.enabled?(:draft_live_versioning)
+
+    made_live_forms.last.json_form_blob
   end
 
   def snapshot
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -2,6 +2,7 @@ class Form < ApplicationRecord
   has_paper_trail
 
   has_many :pages, -> { order(position: :asc) }, dependent: :destroy
+  has_many :made_live_forms
 
   validates :org, :name, presence: true
   def start_page
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -12,17 +12,11 @@ class Form < ApplicationRecord
   def make_live!
     update!(live_at: Time.zone.now)
 
-    made_live_forms.create!(json_form_blob: to_json(include: [:pages])) if FeatureService.enabled?(:draft_live_versioning)
+    made_live_forms.create!(json_form_blob: to_json(include: [:pages]))
   end
 
   def live_version
-    return snapshot unless FeatureService.enabled?(:draft_live_versioning)
-
     made_live_forms.last.json_form_blob
-  end
-
-  def snapshot
-    to_json(include: [:pages])
   end
 
   def name=(val)
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_03_02_161521) do
+ActiveRecord::Schema[7.0].define(version: 2023_03_03_132811) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_02_27_143428) do
+ActiveRecord::Schema[7.0].define(version: 2023_03_02_161521) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -32,6 +32,14 @@ ActiveRecord::Schema[7.0].define(version: 2023_02_27_143428) do
     t.integer "page_order", array: true
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+  end
+
+  create_table "made_live_forms", force: :cascade do |t|
+    t.bigint "form_id"
+    t.json "json_form_blob"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["form_id"], name: "index_made_live_forms_on_form_id"
   end
 
   create_table "pages", force: :cascade do |t|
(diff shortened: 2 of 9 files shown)


db/schema.rb
```diff
@@ -48,5 +48,36 @@
...
+    t.bigint "item_id", null: false
+    t.string "event", null: false
+    t.string "whodunnit"
+    t.jsonb "object"
+    t.datetime "created_at"
+    t.jsonb "object_changes"
+    t.integer "transaction_id"
+    t.index ["item_type", "item_id"], name: "index_versions_on_item_type_and_item_id"
+    t.index ["transaction_id"], name: "index_versions_on_transaction_id"
+  end
```
lfdebrux, line 79: I can't see migrations for these tables, I assume they are from papertrail and not needed for this PR?

db/schema.rb
```diff
@@ -48,5 +48,36 @@
     t.index ["form_id"], name: "index_pages_on_form_id"
   end
 
+  create_table "published_forms", force: :cascade do |t|
```
lfdebrux, line 51: Was chatting to @danworth  about this, he made a good point that any features we add in future might need to be propagated through this table.

I'm not sure it's necessarily a blocking issue or even a major problem, I think we'd only need to regenerate if we make changes to the schema that break backwards-compatibility, but might be worth documenting somewhere so we don't forget about it.

db/schema.rb
aliuk2012: 💯 this is just because I've been switching between `add-paper-trail` branch and this one and haven't migrated up/down each time.

db/schema.rb
aliuk2012: Yeah, this is something @thomasiles and I have been chatting about and thinking at there is still a need for paper-trail if not anything to be able to revert to specific forms or pages.

We will very much have to be mindful of backwards compatibility and its something we kind of "dodged"  a bit when we rolled out the new answer_Settings input_types.

db/migrate/20230217215025_create_published_forms.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePublishedForms < ActiveRecord::Migration[7.0]
+  def change
+    create_table :published_forms do |t|
+      t.references :form, index: true, foreign_key: true
+      t.json :version
```
lfdebrux, line 5: I think this would be better called something like 'snapshot' or 'data', to me it wasn't clear reading the code what 'version' was.

app/models/form.rb
```diff
@@ -1,5 +1,6 @@
 class Form < ApplicationRecord
   has_many :pages, -> { order(position: :asc) }, dependent: :destroy
+  has_many :published_forms
```
lfdebrux, line 3: What's the benefit of calling this `published_forms` instead of `live_forms`?

aliuk2012: I think `live_versions` could be confusing because a dev may assume that form filler will be able to use any record for their submissions when we really only want one live version per form.  This table will have every single live form that could have ended up on GOV.UK. We dont know for sure because its a manual process but there is a possibility that they were "published" on GOV.UK at some point in time. Perhaps if we were only keeping one single record then I would agree with you that `live_version` would seem more appropriate.

db/migrate/20230217215025_create_published_forms.rb
aliuk2012: I see your point and open for suggestions. `version` could be misleading as you may expect a numeric  value rather than json. @alphagov/forms-devs  any other suggestions/preference on what to call this column which will store a copy of form and its pages as JSON?

app/models/form.rb
```diff
@@ -8,10 +9,14 @@ def start_page
 
   def make_live!
     update!(live_at: Time.zone.now)
+
+    published_forms.create!(version: to_json(include: [:pages])) if FeatureService.enabled?(:draft_live_versioning)
```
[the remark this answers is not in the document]
aliuk2012: Let me have a think about your suggestion as the original intention was to eventually remove snapshot method once the feature is enabled. we only using it at the moment to return the current draft form and pages when `live_form` endpoint to called.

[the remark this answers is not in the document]
aliuk2012: I think we will stick with keeping `snapshot` as a temporary method as its only being used in this model class. We dont want it be widely used and then its confusing whether to call `snapshot` or `live_version`

app/models/form.rb
james-sheppard-gov-uk: I'd prefer to reserve use of the word "publish" when we're referring to the activity relating to publishing on GOV.UK. We don't use the word "publish" elsewhere in the team, so following the concept of a ubiquitous language we'd not use it here to prevent possible confusion.

SamJamCul: I think this ties back to what @james-sheppard-gov-uk mentioned about reserving the term 'published' for things that have go through the publishing process for GOV.UK. 

It may be the case that somewhere down the line we would want to distinguish between forms that are live and forms that are currently published (maybe for metrics or something)

DavidBiddle: Even though it's potentially clunky I'd prefer something like `live_forms` or `made_live_forms` here - we've been quite careful to draw the distinction between published and live elsewhere so I think it would be useful to preserve that distinction.

danworth:
I have a general comment about a slight uneasiness about storing the forms/pages as relational models and again as json for the live versions, for example the implications when we run a migration, would it be on both or just the relational model etc. Short of having a crystal ball its difficult to tell if this will actually be a problem or cause greater complexity. The changes here are are far as I can see encapsulated within forms-api behind its api, so if we need to change course in future it shouldn't be too onerous. I think this is a case of it's better to try something and make progress than try to spend too long finding the perfect solution.

I think this also raises interesting questions about the growing use of json to store form data which is something we've long considered and it'll be interesting to see how this plays out.

I guess there are implications for @SamJamCul here, because the live version stored as json would have to include the conditions when they're added?

db/migrate/20230217215025_create_published_forms.rb
SamJamCul: `json_form_blob` or something! It's hard to to come up with something fairly generic as it encapsulates quite a lot.

aliuk2012: `data` or `form_data` might lead you to think that its the form fillers responses. `forms_and_pages_data` blocks us adding other tables into the mix. 

Maybe @SamJamCul suggestion `json_form_blob` is generic enough

app/models/form.rb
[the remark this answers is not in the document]
aliuk2012: @danworth no problem, everything is still behind feature flag because there are still some ui work in forms-admin that needs to be completed before we should be using this.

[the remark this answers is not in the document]
lfdebrux: Makes sense to remove `snapshot` eventually, I just am keen that we avoid duplicate code for as long as we do have the two methods; it would be bad if they diverged accidentally.

db/migrate/20230217215025_create_published_forms.rb
aliuk2012: I've renamed the `versions` column to `json_form_blob`

app/models/form.rb
aliuk2012: I've now renamed `published_forms` table to `made_live_forms`

DavidBiddle (approved):
Looks good 💯

thomasiles (approved):
🥇

lfdebrux (approved):
🎉

PR #210, opened 2023-02-22
Simple routing spike

SamJamCul:

Trello card: https://trello.com/c/P93odeqZ/508-spike-how-to-store-simple-routing-in-the-api

the main doc sets outs the thinking around how we'll introduce routing to the api and runner

### DIFF
--- a/example_routing.md
+++ b/example_routing.md
@@ -90,8 +90,49 @@ Question 1 Condition
 
 In the backend this would look alright, as the condition attached to the question would be `{question_id: 1, match_answer: "Yes", destination_id: 3}`. you know what nevermind!
 
-
 ## Option 3
+
+The `next_page` field could simply be the single source of information for where to calculate the next page. This has the advantage of making it straightforward to identify what the next page should be from a single field. The main issue would be the possibility of overcomplicating a single field, as it would potentially have to store quite a wide range of data (as it would need to effectively implement one of the two previous options *within* the `next_page` field)
+
+### api
+
+```json
+  {
+    id: 1,
+    next_page: {default: 2, conditions: []},
+  }
+```
+
+```json
+  {
+    id: 1,
+    next_page: {
+      default: 2, 
+      conditions: [{question_id: 1, match_answer: "first option", destination_id: 3}]
+    },
+  }
+```
+
+This does turn next page into a good place to store all the information about where the runner should take the user
+
+### runner
+
+The runner could implement a method in the page model that works out the next page in context. 
+
+This way, page objects can have their `next_page` queried, which can return the id for the next page.
+
+```ruby
+  def next_page
+    next_page.conditions.each do |condition|
+      given_answer = session[condition.question_id].answer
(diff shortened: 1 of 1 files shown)


example_routing.md
```diff
@@ -0,0 +1,242 @@
...
+Questions either have a next page when they're unconditional, or a set of conditions if there need to be conditional routing checks. This is not as preferable as the second option, as it requires us to construct more of a branching system whereby all outputs to a conditional question are specified - so all routes out of that question are covered by conditions. The user will implicitly set up routes in different directions, and we'd have to calculate them explicitly. 
+
+There's potentially a bit of value down the line in being extremely explicit about all the ways a route resolves, but it's kind of hard to think about them now.
+
+### api
+```json
+{
+  id: 1,
+  next_page: 2,
+  conditions: nil,
```
DavidBiddle, line 17: Very minor syntax thing but none of these JSON snippets are technically valid - keys need to be strings and `nil` is Ruby:
```suggestion
  "id": 1,
  "next_page": 2,
  "conditions": null,
```

SamJamCul: Ah good spot, no wonder my editor was flagging it as nonsense - I'll tidy it up a bit

DavidBiddle:
We should probably make this an ADR after it's been run past Chris and the tech weekly

example_routing.md
```diff
@@ -0,0 +1,242 @@
...
+Alternatively, question pages will always have a default `next_page` which can be deferred to - but if they have any `conditions`, those will be checked against primarily and take precedence over the `next_page` if any are met. This is what I prefer, as it lets us make minimal additions to questions. 
+
+### api
+
+```json
+{
+  id: 1,
+  next_page: 2,
+  conditions: [
+    {question_id: 1, match_answer: "first option", destination_id: 3},
```
DavidBiddle, line 58: Just clarifying the difference between `page.id` and `page.conditions[0].question_id` - is the idea here that:
- `page.id` determines when this condition gets evaluated in the runner
- `page.conditions[0].question_id` determines the question whose value is being interpreted in the condition

So for this `pages` array:
```json
[
  {
    "id": 1,
    "next_page": 2,
    "conditions": []
  },
  {
    "id": 2,
    "next_page": 3,
    "conditions": [{"question_id": 1, "match_answer": "first option", "destination_id": 4}]
  },
  {
    "id": 3,
    "next_page": 4,
    "conditions": []
  },
  {
    "id": 4,
    "next_page": null,
    "conditions": [],
  }
]
```

The user would:
- complete question 1 and get taken to question 2
- complete question 2, and then depending on their answer to question 1 either get taken to question 3 or question 4

SamJamCul: Yes - because there are conditions in page 2's conditions list, the runner will then check whether the answer to question 1 matches - and if it does, it will use page 4 as the next page

DavidBiddle:
I've started turning this into an ADR here: https://github.com/alphagov/forms/pull/63

PR #212, opened 2023-02-28
Add auditing to our models

aliuk2012:

This is the outcome of https://github.com/alphagov/forms-api/pull/174 where we were initially thinking of using papertrail to snapshot draft/live versions but we ran into issues with associations records and when attempting to delete them.  We decided now to just use Papertrail for auditing purposes and for single record backup/restoration.

- Adds PaperTrail gem
- configures Form/Page to be versioned

This data isn't being surfaced to users at the moment but we can still use this for support queries and eventually it will be surfaced to users when we get to versioning. 

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/db/migrate/20230227143427_create_versions.rb
+++ b/db/migrate/20230227143427_create_versions.rb
@@ -1,0 +1,31 @@
+# This migration creates the `versions` table, the only schema PT requires.
+# All other migrations PT provides are optional.
+class CreateVersions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :versions do |t|
+      t.string   :item_type, null: false
+      t.bigint   :item_id,   null: false
+      t.string   :event,     null: false
+      t.string   :whodunnit
+      t.jsonb :object
+
+      # Known issue in MySQL: fractional second precision
+      # -------------------------------------------------
+      #
+      # MySQL timestamp columns do not support fractional seconds unless
+      # defined with "fractional seconds precision". MySQL users should manually
+      # add fractional seconds precision to this migration, specifically, to
+      # the `created_at` column.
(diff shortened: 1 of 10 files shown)


db/migrate/20230227143427_create_versions.rb
```diff
@@ -0,0 +1,31 @@
+# This migration creates the `versions` table, the only schema PT requires.
```
DavidBiddle, line 1: Shame it's called 'versions' but I guess there's probably not much we can do about it!

DavidBiddle (approved):
Looks good to me! 🎉

db/migrate/20230227143427_create_versions.rb
aliuk2012: @DavidBiddle we can change it, any suggestions? I just wanted to keep the default settings until we start using it more and then tweak it further.

DavidBiddle: Maybe something like `audits`, `audit-versions` or `history` - not sure really, just worried it might be mistaken for something to do with our form versioning. But I don't feel that stongly about it if we can't think of something good!

thomasiles: I think if we use it for all papertail object versioning - as we currently do - we should use `versions` as it's the papertail default and is what people expect across projects.

I did look at [using a custom table/class for form versions](https://github.com/paper-trail-gem/paper_trail#6a-custom-version-classes) and calling it something like `editions`/`revisions`/`audit` as it seems such a fundamental part of the domain. In this case i think using a better name makes sense. 

Doing that makes using papertrail for other models more complicated and it's less flexible, so i'm not sure it's worth it

thomasiles (approved):

PR #227, opened 2023-03-21
New end point to return details of current token being used for the request

aliuk2012:

using an existing active token making  a GET request to `api/v1/access-tokens/caller-identity` will return the details for token that's being used in the existing request.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/api/v1/access_tokens_controller.rb
+++ b/app/controllers/api/v1/access_tokens_controller.rb
@@ -27,7 +27,7 @@ class Api::V1::AccessTokensController < ApplicationController
 
   def caller_identity
     if @access_token
-      render json: @access_token.as_json(except: [:token_digest]).to_json, status: :ok
+      render json: @access_token.to_json, status: :ok
     else
       render json: { error: "Not found - No token used." }.to_json, status: :not_found
     end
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/access_tokens_controller.rb
```diff
@@ -25,6 +25,14 @@ def deactivate
     render json: { status: "`#{@access_token.owner}` has been deactivated" }.to_json, status: :ok
   end
 
+  def caller_identity
+    if @access_token
+      render json: @access_token.as_json(except: [:token_digest]).to_json, status: :ok
```
danworth, line 30: I can't see any harm in returning the digest for what its worth (its derived from the key they have). During the initial debugging of this new feature we needed to see the digest to see it changing on each invocation, that we got from querying the db directly since its also not returned in the get all endpoint. Going forward it might not be as useful so we could add it if necessary 👍

danworth (approved):
Nice thanks Al, should make future work to script managing keys simpler 🏅

app/controllers/api/v1/access_tokens_controller.rb
aliuk2012: Good point, I've included the token_digest. I excluded it at first because I thought it might be a security risk but if the token being passed in is already known then showing the digest along with the all the other info shouldn't be necessary.

PR #228, opened 2023-03-21
Return a list of forms across all organisations

aliuk2012:
The problem with making the org param required is that it stops us from searching for all forms across all organisations.

If you pass in `org` as a query param (which `forms-admin` and `forms-runner` already do), it will return a list of forms for that organisation. If you leave off the query param it will return a list of forms.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


danworth (approved):
Looks good, pretty sure we pass org as a param currently so this won't break any existing calls to forms-api 🏅

PR #230, opened 2023-03-22
Add Trello Poster

lfdebrux:
Make it so that if we add a link to a Trello card in a PR description, that PR will be attached to the Trello card if it isn't already.

This should make linking Trello cards to PRs slightly easier.

Trello card: https://trello.com/c/fQNBNRqv

### DIFF
--- a/.github/workflows/trello-poster.yml
+++ b/.github/workflows/trello-poster.yml
@@ -1,0 +1,14 @@
+name: Trello Poster
+on:
+  pull_request:
+    types: [created, edited]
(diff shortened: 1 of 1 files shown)


DavidBiddle (approved):
Looks good to me! 🎉 

Out of interest, will this work for private/internal repos as well as public ones?

lfdebrux:
> Out of interest, will this work for private/internal repos as well as public ones?

It should do! Although we do need to add it to each repo individually.

PR #231, opened 2023-03-24
Create conditions table and model

SamJamCul:

Trello card: https://trello.com/c/FbZxLSWp/561-api-migration-for-simple-routing

This adds a Conditions table, which contains a set of references to the Pages table. 

There's a bit of complexity around which references are necessary and which are optional. The idea is that a Condition has to be attached to a specific page, which is the `routing_page`. This is a mandatory relationship, and deleting the page destroys this condition. 

There are also two other page references, one to the page whose answer will be checked, the `check_page`, and the other to whom the user will skip to if the condition is met, the `goto_page`. If the user deletes the associated page, the reference is simply removed, while the condition remains.

Updating comment for trello testing!

31/03/23 **Update**: I've added a commit to add an additional column which keeps track of whether the condition should skip _past_ the last page of the form, and bring the user to the check your answers page. We can't track this using the `goto_page` foreign key column, as there isn't a page object to key to! We did consider making use of the fact that the `goto_page` column could be `null` which could represent skipping to the last page, but then we have a bit of an issue with deciding whether a user has actively set the condition to skip to the last page, or if they just didn't notice that they skip to the last page by default.

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -1,0 +1,7 @@
+class Condition < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :routing_page, class_name: "Page"
+  belongs_to :check_page, class_name: "Page", optional: true
+  belongs_to :goto_page, class_name: "Page", optional: true
+end
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -2,6 +2,9 @@ class Page < ApplicationRecord
   has_paper_trail
 
   belongs_to :form
+  has_many :routing_conditions, class_name: "Condition", foreign_key: "routing_page_id", dependent: :destroy
+  has_many :check_conditions, class_name: "Condition", foreign_key: "check_page_id", dependent: :nullify
+  has_many :goto_conditions, class_name: "Condition", foreign_key: "goto_page_id", dependent: :nullify
   acts_as_list scope: :form
 
   ANSWER_TYPES = %w[number address date email national_insurance_number phone_number selection organisation_name text name].freeze
--- a/db/migrate/20230324091733_create_conditions_table.rb
+++ b/db/migrate/20230324091733_create_conditions_table.rb
@@ -1,0 +1,11 @@
+class CreateConditionsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :conditions do |t|
+      t.references :check_page
+      t.references :routing_page
+      t.references :goto_page
+      t.string :answer_value
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20230324091733_create_conditions_table.rb
+++ b/db/migrate/20230324091733_create_conditions_table.rb
@@ -1,9 +1,9 @@
 class CreateConditionsTable < ActiveRecord::Migration[7.0]
   def change
     create_table :conditions do |t|
-      t.references :check_page
-      t.references :routing_page
-      t.references :goto_page
+      t.references :check_page, comment: "The question page this condition looks at to compare answers"
+      t.references :routing_page, comment: "The question page at which this conditional route takes place"
+      t.references :goto_page, comment: "The question page which this conditions will skip forwards to"
       t.string :answer_value
       t.timestamps
     end
--- a/spec/factories/conditions.rb
+++ b/spec/factories/conditions.rb
@@ -1,0 +1,8 @@
+FactoryBot.define do
+  factory :condition do
+    routing_page { build :page }
+    check_page { nil }
+    goto_page { nil }
+    answer_value { nil }
+  end
+end
--- a/spec/factories/conditions.rb
+++ b/spec/factories/conditions.rb
@@ -4,5 +4,6 @@ FactoryBot.define do
     check_page { nil }
     goto_page { nil }
     answer_value { nil }
+    skip_to_end { false }
   end
 end
--- a/db/migrate/20230331141402_add_skip_to_end_to_conditions.rb
+++ b/db/migrate/20230331141402_add_skip_to_end_to_conditions.rb
@@ -1,0 +1,5 @@
+class AddSkipToEndToConditions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :conditions, :skip_to_end, :boolean, default: false
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -24,6 +24,18 @@ ActiveRecord::Schema[7.0].define(version: 2023_03_24_120801) do
     t.string "description"
   end
 
+  create_table "conditions", force: :cascade do |t|
+    t.bigint "check_page_id"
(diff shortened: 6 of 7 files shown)


app/models/condition.rb
```diff
@@ -0,0 +1,7 @@
+class Condition < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :routing_page, class_name: "Page", foreign_key: "routing_page_id"
+  belongs_to :check_page, class_name: "Page", optional: true, foreign_key: "check_page_id"
+  belongs_to :goto_page, class_name: "Page", optional: true, foreign_key: "goto_page_id"
```
SamJamCul, line 6: The idea here is that for a Condition object, you can call `c.routing_page` to find out which page it's attached to, `c.check_page` to find the page that it's checking, and `c.goto_page` to find the page that it skips to

app/models/page.rb
```diff
@@ -2,6 +2,9 @@ class Page < ApplicationRecord
   has_paper_trail
 
   belongs_to :form
+  has_many :routing_conditions, class_name: "Condition", foreign_key: "routing_page_id", dependent: :destroy
+  has_many :check_conditions, class_name: "Condition", foreign_key: "check_page_id", dependent: :nullify
+  has_many :goto_conditions, class_name: "Condition", foreign_key: "goto_page_id", dependent: :nullify
```
SamJamCul, line 7: Here there are a set of associations for page objects to conditions - you can find all the conditions taking place on a given page with `p.routing_conditions`, all the conditions that check this page with `p.check_conditions` and all the conditions that direct towards this page with `p.goto_conditions` (I'm definitely open to naming suggestions to make this less confusing)

I did initially set it up so that you could do `p.conditions` to return associated conditions, but it's really messy when you're getting back conditions that can be referring to this page in different capacities. It's something we could add if it turns out it's valuable though.

spec/factories/conditions.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :condition do
+    routing_page { build :page }
+    check_page { nil }
+    goto_page { nil }
+    answer_value { nil }
```
SamJamCul, line 6: I don't know if I should build a more full-featured factory here - the conditions model is relatively simple, so I don't think we need to over-egg the testing in forms-api

app/models/condition.rb
```diff
@@ -0,0 +1,7 @@
+class Condition < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :routing_page, class_name: "Page"
+  belongs_to :check_page, class_name: "Page", optional: true
+  belongs_to :goto_page, class_name: "Page", optional: true
```
SamJamCul, line 6: he idea here is that for a Condition object, you can call c.routing_page to find out which page it's attached to, c.check_page to find the page that it's checking, and c.goto_page to find the page that it skips to

db/migrate/20230324091733_create_conditions_table.rb
lfdebrux, line 1: My apologies, you might need to rename this migration, as I just merged #232 which has some migrations with more recent datetimes.

spec/factories/conditions.rb
thomasiles: I think keep it simple like this if you can, we found the factory for form got a bit complicated when we were adding more traits to cover other scenarios

db/migrate/20230324091733_create_conditions_table.rb
```diff
@@ -0,0 +1,11 @@
+class CreateConditionsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :conditions do |t|
+      t.references :check_page
```
thomasiles, line 4: is it useful add `:comment` to each of the references describing what role it plays (e.g. from your PR intro)

for example: 
```t.references :check_page, comment: "The page this condition is attached to"```

We don't do that in other places but it might make things a bit clearer for those infra devs mucking around in the db

File: app/models/condition.rb [hunk 6]
```diff
@@ -0,0 +1,7 @@
+class Condition < ApplicationRecord
+  has_paper_trail
```
thomasiles, line 2: This is fine, I think.

This means all changes will be recorded in the version table. They will be scoped to the Condition model so shouldn't interfere with any other draft/live features. If that's not useful, we can delete/ignore them.

The draft/live work has add two new routes `/draft` and `/live`, which the runner uses. These add pages into the main forms json so provide a complete description of a form. We'll need to include the conditions in these routes. When you get to that bit give us a shout if you need help.

thomasiles (approved):
Nice one!🏅 This looks nice and simple to me. It's a bit hard to get a handle on what the three page references do, but that's probably because they are new.

db/migrate/20230324091733_create_conditions_table.rb
DavidBiddle: i think we concluded at the weekly yesterday that this should be OK without being renamed. I believe we will need to resolve the conflict in schema.rb by rebasing this branch onto main and running the migrations though

db/migrate/20230324091733_create_conditions_table.rb
DavidBiddle: This is a nice idea I think!

DavidBiddle (approved):

app/models/condition.rb
DavidBiddle: @SamJamCul These differ slightly from the [equivalent names in the ADR](https://github.com/alphagov/forms/pull/65/files#diff-8bcea9845a999e062259cd5f1a83bb9412e41bb0a60b6d937eab0d717d8c06adR15-R18), so we should make sure we document these somewhere

app/models/page.rb
DavidBiddle: Don't think we need to change anything in this PR, but just noting that we'll have to do some eager loading to make these associations available in the API/Admin

app/models/condition.rb
SamJamCul: I don't know if we need to be that tethered to what's in the ADR. The ADR is meant to capture the decision to build the routing work as a table in our DB; the way we implement it exactly isn't something we can have predicted in the ADR. I think this PR could be linked as part of the ADR somewhere, to show that this work was following on from it?

app/models/condition.rb
SamJamCul: Do you think we should start adding conditions to the live form jsons as part of this PR, or save it for further down the line?

app/models/condition.rb
DavidBiddle: Yeah that makes sense I think 👍

SamJamCul:
@thomasiles @DavidBiddle I've added another commit which adds a `skip_to_end` column, which will be used to track conditions that skip to the end of the form rather than a specific question.

PR #232, opened 2023-03-24
Refactors to draft/live functionality and associated tests

lfdebrux:
While working on https://trello.com/c/845sNCtW I found there were a bunch of things that it would be good to clean up in our models to make it easy to tell when a form had a draft version and/or a live version.

I've split those changes out into a separate PR, to break things up a bit.

- It was hard to write model specs that change a form or a page while doing the same bookkeeping as the controllers, we move that business logic into a new `FormUpdateService`
- The datetime stamps for live_at and made_live_forms.last weren't consistent with the updated_at of the form, so we couldn't compare reliably
- The logic for creating a JSON form blob was scattered all over in our code and in our specs, we restore the `snapshot` method to bring it all into one place

Those are the major ones, there are also some small tweaks I snuck in as well ;)

### DIFF
--- a/app/controllers/api/v1/pages_controller.rb
+++ b/app/controllers/api/v1/pages_controller.rb
@@ -9,8 +9,7 @@ class Api::V1::PagesController < ApplicationController
   def create
     new_page = form.pages.new(page_params)
 
-    if new_page.save
-      form.update!(question_section_completed: false)
+    if new_page.save_and_update_form
       render json: { id: new_page.id }, status: :created
     else
       render json: new_page.errors.to_json, status: :bad_request
@@ -24,8 +23,7 @@ class Api::V1::PagesController < ApplicationController
   def update
     page.assign_attributes(page_params)
 
-    if page.save
-      form.update!(question_section_completed: false)
+    if page.save_and_update_form
       render json: { success: true }.to_json, status: :ok
     else
       render json: page.errors.to_json, status: :bad_request
@@ -33,8 +31,7 @@ class Api::V1::PagesController < ApplicationController
   end
 
   def destroy
-    page.destroy!
-    form.update!(question_section_completed: false)
+    page.destroy_and_update_form!
     render json: { success: true }.to_json, status: :ok
   end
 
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/pages_controller.rb
```diff
@@ -24,7 +24,7 @@ def show
   def update
     page.assign_attributes(page_params)
 
-    if page.save
+    if FormUpdateService.save_page(page)
```
aliuk2012: Not sure I follow the difference here,

app/models/form.rb
```diff
@@ -20,6 +20,10 @@ def draft_version
     snapshot.to_json
   end
 
+  def live_at
+    return made_live_forms.last.created_at
+  end
```
aliuk2012: Could we add a test for this?

app/service/form_update_service.rb
```diff
@@ -0,0 +1,23 @@
...
+      form.save
+    end
+
+    def update(form, **options)
+      form.update(**options)
+    end
+
+    def destroy!(form)
+      form.destroy!
+    end
```
aliuk2012, line 13: not sure I'm following this service. It feels like we just trying to duplicate what ActiveRecord already gives us.

lfdebrux: @thomasiles @aliuk2012 and I discussed and agreed a different approach, see [64290fa](https://github.com/alphagov/forms-api/pull/232/commits/64290fa4f14a8225abe7319025a7eaddc68682a5).

app/models/form.rb
lfdebrux: Done!

app/controllers/api/v1/pages_controller.rb
lfdebrux: @thomasiles @aliuk2012 and I discussed and agreed a different approach, see [f1061ed](https://github.com/alphagov/forms-api/pull/232/commits/f1061edc2712004b5f1532e8982fae9bf901c3bd).

aliuk2012 (approved):
Some good refactors and I like the reintroduction of "snapshots". 👍

PR #249, opened 2023-04-13
Add conditions controller

SamJamCul:

This adds a set of endpoints for the conditions table. It makes use of the rails resource system, so it has all the standard routes. 

Trello card: https://trello.com/c/Y5noSn8H/660-add-conditions-endpoint-to-api

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -4,4 +4,14 @@ class Condition < ApplicationRecord
   belongs_to :routing_page, class_name: "Page"
   belongs_to :check_page, class_name: "Page", optional: true
   belongs_to :goto_page, class_name: "Page", optional: true
+
+  has_one :form, through: :routing_page
+
+  def save_and_update_form
+    save && form.update!(question_section_completed: false)
+  end
+
+  def destroy_and_update_form!
+    destroy! && form.update!(question_section_completed: false)
+  end
 end
--- a/spec/request/api/v1/conditions_controller_spec.rb
+++ b/spec/request/api/v1/conditions_controller_spec.rb
@@ -13,7 +13,7 @@ describe Api::V1::ConditionsController, type: :request do
       let(:routing_page) { create :page, form: }
 
       it "returns 200 and an empty json array" do
-        get "/api/v1/forms/#{form.id}/pages/#{routing_page.id}/conditions", as: :json
+        get form_conditions_path(form, routing_page), as: :json
         expect(response.status).to eq(200)
         expect(response.headers["Content-Type"]).to eq("application/json")
         expect(json_body).to eq([])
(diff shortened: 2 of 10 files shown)


spec/request/api/v1/conditions_controller_spec.rb
```diff
@@ -0,0 +1,312 @@
...
+#         expect(response.headers["Content-Type"]).to eq("application/json")
+#         expect(json_body).to eq({ success: 1 })
+#       end
+
+#       it "does not change order of pages" do
+#         get "/api/v1/forms/#{form_with_pages.id}/pages"
+#         expect(json_body.map { |p| p[:id] }).to eq([page_to_move.id, second_page.id, third_page.id, fourth_page.id, last_page.id])
+#       end
+#     end
+#   end
```
aliuk2012, line 311: Should all these tests be commented out?

aliuk2012, line 311:
```suggestion
```

aliuk2012 (approved):

app/models/condition.rb
```diff
@@ -4,4 +4,15 @@ class Condition < ApplicationRecord
...
+
+  has_one :form, through: :routing_page
+
+  def save_and_update_form
+    save && form.update!(question_section_completed: false)
+  end
+
+  def destroy_and_update_form!
+    form = self.form
+    destroy! && form.update!(question_section_completed: false)
```
aliuk2012, line 16:
```suggestion
    destroy! && form.update!(question_section_completed: false)
```

spec/request/api/v1/conditions_controller_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+  let(:goto_page) { create :page, form: }
+  let(:condition) { create :condition }
+  let(:condition_id) { condition.id }
+
+  describe "#index" do
+    context "when no conditions exist for a page" do
+      let(:routing_page) { create :page, form: }
+
+      it "returns 200 and an empty json array" do
+        get "/api/v1/forms/#{form.id}/pages/#{routing_page.id}/conditions", as: :json
```
aliuk2012, line 16: all GET/PUTS etc paths should be using rails url helpers. I think the pages request specs need to be updated. forms request specs have been updated to use `xxxx_path` but pages hasn't when these request specs were first written we hadn't migrated over to rails so we couldn't use the url helpers until after the migration.

aliuk2012: I've created this PR to fix the pages request specs https://github.com/alphagov/forms-api/pull/250

app/models/condition.rb
SamJamCul: removed it!

PR #258, opened 2023-04-27
Add routing warnings - Go To page doesn't exist

aliuk2012:

This is the first time that `forms-api` will be returning custom errors for the client to handle.  This PR adds a new method called `validation_errors` which dynamically evaluates the data when it is called to check if the route is valid. If it is not then it returns an object with a shorthand human readable error. `forms-admin` will use this value to look up the full user friendly error description. We did this because `forms-api` would not know what language the `forms-admin` is using without us also including the language in every request. 

```json
  {
      "id": 33,
      "check_page_id": 4,
      "routing_page_id": 4,
      "goto_page_id": 12,
      "answer_value": "Cat",
      "created_at": "2023-04-25T09:14:55.978Z",
      "updated_at": "2023-04-28T10:49:27.283Z",
      "skip_to_end": false,
      "validation_errors": [
          {
              "name": "goto_page_doesnt_exist"
          }
      ]
  }

```

Trello card: https://trello.com/c/tcKng6uH/694-add-go-to-page-does-not-exist-error-to-conditions-in-api

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -14,4 +14,16 @@ class Condition < ApplicationRecord
   def destroy_and_update_form!
     destroy! && form.update!(question_section_completed: false)
   end
+
+  def validation_errors
+    [warning_goto_page_doesnt_exist].compact
+  end
+
+  def warning_goto_page_doesnt_exist
+    page = form.pages.find_by(id: goto_page_id)
+    return nil if page.present?
+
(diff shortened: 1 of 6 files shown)


DavidBiddle (approved):
Looks good! 🎉

PR #262, opened 2023-05-03
Fix live_at Time

danworth:
Specifies the order for made_live_forms so that the ordering and selection of the latest version with ActiveRecord.last function correctly selects the latest version based on the created_at time. Without this the ordering is inconsistent because the resulting `SELECT` statement does not include an `ORDER BY` clause; postgres will varying the order of returned rows without an `ORDER BY` clause due to the `synchronize_seqscans` optimisation.

Updates the integration tests to create multiple live versions and ensure the returned `live_at` is based upon the created_at time of the latest version.

### NOTES ###
Tested this locally and by deploying the branch to AWS dev. The `live_at` times are now consistent and consistently reflect the latest live version (note that this inconsistency was also present against the same data base but was more prevalent when comparing results form different databases e.g. PaaS vs AWS). 

The resulting SQL statement correctly includes the `ORDER BY` clause
```
  MadeLiveForm Load (3.1ms)  SELECT "made_live_forms".* FROM "made_live_forms" WHERE "made_live_forms"."form_id" = $1 ORDER BY "made_live_forms"."created_at" ASC  [["form_id", 144]]
```

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -2,7 +2,7 @@ class Form < ApplicationRecord
   has_paper_trail
 
   has_many :pages, -> { order(position: :asc) }, dependent: :destroy
-  has_many :made_live_forms, dependent: :destroy
+  has_many :made_live_forms, -> { order(created_at: :asc) }, dependent: :destroy
 
   validates :org, :name, presence: true
   def start_page
--- a/spec/models/form_spec.rb
+++ b/spec/models/form_spec.rb
@@ -148,9 +148,13 @@ RSpec.describe Form, type: :model do
       expect(form.live_at).to be_nil
     end
 
-    it "returns the time when the form was made live" do
-      made_live_form = create :made_live_form
-      expect(made_live_form.form.live_at).to eq made_live_form.created_at
+    it "returns the created_at time of the latest live version" do
+      form = create :form
+      _first_live_version = form.make_live!
+      _second_live_version = form.make_live!
+      third_live_version = form.make_live!
+
+      expect(form.live_at).to eq third_live_version.created_at
     end
   end
 


app/models/form.rb
```diff
@@ -2,7 +2,7 @@ class Form < ApplicationRecord
   has_paper_trail
 
   has_many :pages, -> { order(position: :asc) }, dependent: :destroy
-  has_many :made_live_forms, dependent: :destroy
+  has_many :made_live_forms, -> { order(created_at: :asc) }, dependent: :destroy
```
lfdebrux, line 5: I'm wondering about the timing implications of this... might be premature optimisation, but could we use the primary key instead of `created_at`?

danworth: I'm weary of using primary keys for this ordering because although that would order them in some semblance of insertion (and be quicker query) I don't think we necessarily want that, for example [`make_live` accepts a `live_at`  time](https://github.com/alphagov/forms-api/blob/main/app/models/form.rb#L12) which is then set as the `created_at` time on the live version. This implies that the order needs to be by `created_at` since we could insert a new version with a higher primary key yet with an earlier `created_at`. Also using the primary key for sorting makes this kind of data migration harder because the keys have to align to guarantee the same order. I think it makes sense to use something within our domain model to control the order rather than table primary keys.

I think `created_at` is a bit ambiguous to me, but it seems to be taken as the time the version was created, rather than the time the row was created in the database given how we set it ourselves. I'm not as close to this domain model but from what I understand the `created_at` seems to be the attribute to order on. If query performance suffers we can add an index to the `created_at` column?

lfdebrux: Yeah, the `created_at` time is supposed to represent the time the version was made live, so it is the right thing to order on. I take your point about `make_live!` accepting a time, I had forgotten that. In our domain model we were expecting that rows in the `made_live_forms` would be immutable and only be created with `live_at` at set to now in the future, but we haven't enforced that.

I understand about the option about making an index, at this point without much in the way of performance metrics it's hard to tell though whether that would help or hinder right? I guess though in this case since we're going to be inserting rarely we probably don't need to worry about slowdown due to that...

lfdebrux: I'll approve this PR as-is and maybe we can create a tech debt ticket or something to look at this again in future.

lfdebrux (approved):

app/models/form.rb
danworth: Thanks Laurence, this was an interesting read https://www.postgresql.org/docs/current/indexes-ordering.html. I was curious so I had a noodle on a local postgres instance.
After creating an index on `created_at` 
```
forms-api=# CREATE INDEX index_made_live_forms_on_created_at ON made_live_forms USING BTREE (created_at ASC NULLS LAST);
```

postgres won't use it, instead preferring to do a separate sort by operation
```
forms-api=# EXPLAIN SELECT * FROM made_live_forms WHERE form_id = 144 ORDER BY created_at ASC;
                               QUERY PLAN
-------------------------------------------------------------------------
 Sort  (cost=10.67..10.68 rows=3 width=1393)
   Sort Key: created_at
   ->  Seq Scan on made_live_forms  (cost=0.00..10.65 rows=3 width=1393)
         Filter: (form_id = 144)
(4 rows)
```

if I force postgres to use it with `forms-api=# set enable_sort=off;` then it does use the index but the total cost is higher (the second number after the `..`)
```
forms-api=# EXPLAIN SELECT * FROM made_live_forms WHERE form_id = 144 ORDER BY created_at ASC;
                                                  QUERY PLAN
---------------------------------------------------------------------------------------------------------------
 Index Scan using index_made_live_forms_on_created_at on made_live_forms  (cost=0.14..24.16 rows=3 width=1393)
   Filter: (form_id = 144)
(2 rows)
```

I'm by no means an expert on this and it was on a local instance of postgres which might not have the parameters or setup as our AWS/PaaS servers but I think its ok to go with as-is and keep an eye on query performance later on.

lfdebrux: Awesome work, thanks for doing that! That is a really interesting result.

PR #264, opened 2023-05-04
Align Github Workflow Across Apps

danworth:
This commit aligns the ruby on rails workflow with out other apps. The matrix testing of Ruby versions was not being used since it contained a single version and it is simpler to use the standard ruby/setup-ruby workflow which will install the version of ruby in the .ruby-version file.

Makes our CI workflow for repo apps consistent between our apps.
See also: 
https://github.com/alphagov/forms-admin/pull/417
https://github.com/alphagov/forms-runner/pull/311


aliuk2012 (approved):

.github/workflows/test.yml
```diff
@@ -36,24 +29,22 @@ jobs:
     env:
       ENV: "test"
```
lfdebrux, line 30: I wonder why this isn't `RAILS_ENV` 🤔 I guess it works though?

lfdebrux (approved):
Big improvement, thanks

PR #269, opened 2023-05-10
Warn users if a route that is the next page

aliuk2012:

The API needs to calculate whether the page being directed to is going to result in at least one question page being skipped - otherwise, the route will, regardless of the condition being triggered, simply go on to the next page. This isn’t a breaking error necessarily, but it does represent a trivial condition being in place

Trello card: https://trello.com/c/K57UF6bv/738-add-routing-goes-to-the-next-page-error-to-conditions-in-api

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -19,6 +19,7 @@ class Condition < ApplicationRecord
     [
       warning_goto_page_doesnt_exist,
       warning_answer_doesnt_exist,
+      warning_routing_to_next_page,
     ].compact
   end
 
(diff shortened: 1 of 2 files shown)


app/models/condition.rb
```diff
@@ -36,6 +37,17 @@ def warning_answer_doesnt_exist
     { name: "answer_value_doesnt_exist" }
   end
 
+  def warning_routing_to_next_page
+    return nil if check_page.nil? || goto_page.nil?
+
+    routing_page_position = check_page.position
+    goto_page_position = goto_page.position
+
+    return { name: "cannot_route_to_next_page" } if goto_page_position == (routing_page_position + 1)
```
SamJamCul, line 46: Do we have a different warning that we'll use for routing to the CYA from the final page? It's possible that we could incorporate that check into this warning - or build another warning that can handle stuff to do with CYA routes

SamJamCul (approved):
Looking good, just a question about the CYA version of this

app/models/condition.rb
aliuk2012: Good question, I think this is something we will need to pick up when we add CYA as an option to the goto pages. CYA is very different from other pages as the concept doesn't really exist in our data model, this is why I think we need to pick it up as another ticket.

PR #270, opened 2023-05-10
Warn users if their route ends up creating an endless loop

aliuk2012:

The API needs to calculate the whether the routing for a given condition makes sense from a sequencing perspective. That is, the question that’s being checked takes place before the question that’s being directed to. This should only happen when a user reorders questions.

Trello card: https://trello.com/c/iFDqUmDc/739-add-question-routing-is-out-of-sequence-error-to-conditions-in-api

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


DavidBiddle (approved):
Looks good to me :tada:

PR #283, opened 2023-06-02
Add creator to form

jamie-o-wilkinson:
We want to be able to give new users a ‘trial’ role so they have permissions to do a limited number of things while deciding whether GOV.UK Forms is a good fit for their team.

In order for a user of forms-admin with the trial user to see only forms they have created, we need to add creator data to the form model in forms-api, and allowing querying by creator.

Trello card:
https://trello.com/c/Bfl4ckBp

#### Checklist

- [N/A] I've updated the documentation in (If any documentation requires updating)
  - [N/A] README.md
  - [N/A] Elsewhere (please link)


app/controllers/api/v1/forms_controller.rb
```diff
@@ -1,8 +1,11 @@
 class Api::V1::FormsController < ApplicationController
   def index
     org = params[:org]
+    creator = params[:creator]
     forms = if org.present?
               Form.where(org:)
+            elsif creator.present?
+              Form.where(creator:)
```
lfdebrux, line 8: With the api we're presenting, where you can have query params `creator` or `org`, I'd kind of expect that doing a query `/forms?org=test-org&creator=test-user` would give me forms that were created by the user in that org. Is that something we can support?

aliuk2012: I'm not keen on how we currently structure this index endpoint. Its a pity we dont have orgs stored in forms-api then we could do `/organisation/:org_id/forms` to return all forms for an organisation. 

for trial users, I'm assuming they wouldn't have an org? either way maybe some of the restricting if they should see the forms should be in the forms-admin form pundit policy.

jamie-o-wilkinson: Yeah that does sound neater for org access @aliuk2012. How do you imagine accessing all forms created by a given user would work alongside that?

Currently (on my branch) the form-policy in forms-admin checks for if the current user has the trial role, and then queries by creator instead of by org.

jamie-o-wilkinson: @lfdebrux I think it would be another (nested) condition in the logic for that endpoint. It does start to get a little messy at that point though

jamie-o-wilkinson: @lfdebrux and @aliuk2012 I've updated this slightly so that it supports filtering by `org` and/or `creator_id` using scopes defined on the Form model. I thought this looked cleaner than adding nested logic to the `index` action. Let me know what you think!

I decided not to go down the route of the wider refactor of using different endpoints as we discussed @aliuk2012 since at this point we still only have the 2 fields we're querying by and felt like it might be overkill for this bit of work, though if we add a concept of "team" in the future it might be a good time.

app/controllers/api/v1/forms_controller.rb
```diff
@@ -1,11 +1,11 @@
...
   def index
     org = params[:org]
-    forms = if org.present?
-              Form.where(org:)
-            else
-              Form
-            end
+    creator_id = params[:creator_id]
+
+    forms = Form.where(nil)
```
lfdebrux, line 6: I don't understand this 😅 Is this the same as `Form.all`?

app/controllers/api/v1/forms_controller.rb
```diff
@@ -1,11 +1,11 @@
...
+    forms = forms.filter_by_org(org) if org.present?
+    forms = forms.filter_by_creator_id(creator_id) if creator_id.present?
```
lfdebrux, line 8: This is very neat! 🙌

lfdebrux (approved):

app/controllers/api/v1/forms_controller.rb
jamie-o-wilkinson: Yes, yes it is! I'm using that instead now 🙃

lfdebrux (approved):

PR #284, opened 2023-06-02
Delete conditions when changing answer settings

SamJamCul:
We want to ensure that conditions that check against selection questions only exist so long as only one option can be selected. It's currently not possible to set up conditions against selections that can have multiple options, so if that change is made, conditions relying on the page will be destroyed.

This also refactors deleting conditions when the answer type changes, so that the check_conditions are removed. This is currently identical to removing the routing_conditions, but in the future we may differentiate between them.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


spec/models/page_spec.rb
```diff
@@ -70,25 +71,43 @@
...
+        it "deletes any conditions" do
+          page.answer_settings["only_one_option"] = "0"
+          page.save_and_update_form
+          expect(page.reload.check_conditions).to be_empty
+        end
+      end
+
+      context "when the answer settings change while still restricting to only one option" do
+        it "does not delete any conditions" do
+          page.question_text = "what is going on?"
```
aliuk2012, line 108: Isn't this change the question text not the answer settings options?

SamJamCul: sorted, thanks for the suggestion!

PR #286, opened 2023-06-09
Prevent error when 'None of the above' is the routing answer

DavidBiddle:
Stops a validation error appearing incorrectly when 'None of the above' is the routing answer. The error will still appear if "None of the above' is not enabled as an option in the seleciton settings page.

Trello card: https://trello.com/c/uqnosctc/826-choosing-none-of-the-above-answered-as-should-not-display-a-validation-error

#### Checklist

  - [n/a] README.md


jamie-o-wilkinson (approved):
Looks good 🎉

PR #290, opened 2023-06-19
Add shell command trace to container run command

lfdebrux:
It is useful when looking at startup logs to be able to know what command was run, to understand the context of the log outputs.

See https://github.com/alphagov/forms-admin/pull/498 for more details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,4 +47,4 @@ COPY --chown=ruby:ruby --from=build /app /app
 
 EXPOSE 9292
 
-CMD ["/bin/sh", "-c", "rake db:migrate && rails s -b 0.0.0.0 -p 9292"]
+CMD ["/bin/sh", "-o", "xtrace", "-c", "rake db:migrate && rails s -b 0.0.0.0 -p 9292"]


Dockerfile
```diff
@@ -47,4 +47,4 @@ COPY --chown=ruby:ruby --from=build /app /app
 
 EXPOSE 9292
 
-CMD ["/bin/sh", "-c", "rake db:migrate && rails s -b 0.0.0.0 -p 9292"]
+CMD ["/bin/sh", "-o", "xtrace", "-c", "rake db:migrate && rails s -b 0.0.0.0 -p 9292"]
```
danworth, line 50: I'm a little bit wary of this because its like leaving a debug statement in production code.  Although its probably unlikely I'm not certain that either now or in the future that there isn't some exceptional scenario in which xtrace could spit something out to the logs that we wouldn't want it to but perhaps I'm being overly cautious.

I wonder if there are other ways to address the visibility or understanding of what happens when our apps start? I think the essential thing to know is that it runs the migration and then starts the server as it does on PaaS today. 

Separately to this, we need to address the issue of two tasks starting and trying to run the migration at the same time (the last one to begin is denied, exits, then replaced which is unideal) and we might end up with a start.sh style script which would we could `echo` out some help logs.

lfdebrux: > I'm a little bit wary of this because its like leaving a debug statement in production code. Although its probably unlikely I'm not certain that either now or in the future that there isn't some exceptional scenario in which xtrace could spit something out to the logs that we wouldn't want it to but perhaps I'm being overly cautious.

Hmm, I see what you're saying, but it does seem unlikely... What sort of things do we not want logged?

> I wonder if there are other ways to address the visibility or understanding of what happens when our apps start? I think the essential thing to know is that it runs the migration and then starts the server as it does on PaaS today.

So for me I thought this would help in making it absolutely clear what commands ran for a particular execution. When we were debugging issues with the migration it was hard to know if the migration command was running as expected, because there is no output if there is no migration to run. So while we knew what command to expect from reading the Dockerfile, I was wondering if we was looking at the right file (or the right version of the file), or if the command is being overridden, or even if the command did run but the logs were missing some of the output...

> Separately to this, we need to address the issue of two tasks starting and trying to run the migration at the same time (the last one to begin is denied, exits, then replaced which is unideal) and we might end up with a start.sh style script which would we could echo out some help logs.

I was thinking that once we were on AWS fully this would be something we would deal with in the deploy pipeline, rather than in the app definition. That seems simpler to me, what do you think?

danworth (approved):
I'm probably being overly cautious due to seeing -x causing issues when left in elsewhere. We'll need to keep it in mind should the command change. If this helps people more quickly understand what's happening from the logs then it seems like a positive change. I'll go through the container start up process and how to figure out what version of code is in any environment with the team too.

Thanks for your patients Laurence 👍

PR #293, opened 2023-06-29
Make update_org_for_creator endpoint return status

jamie-o-wilkinson:
By default `update_org_for_creator` returns 204 No Content, but to be consistent with other endpoints we should return 200 until we have a review of what status codes we should be returning for endpoints that conventionally would return no body in the response.

Trello card: https://trello.com/c/4Lxkorfb


lfdebrux (approved):

lfdebrux:
Thanks for doing this. We should review with the rest of the dev team whether we actually need to return these `{ success: true }` bodies when 204 No Content would suffice, but let's be consistent for now.

jamie-o-wilkinson:
Trello card to review status codes here: https://trello.com/c/alZu2RlP

PR #295, opened 2023-07-05
Add organisation_id to form model

lfdebrux:
### What problem does the pull request solve?

We want to associate forms with organisations using a stable key, however organisations slugs can change, so we want to change the key to be the primary key of the organisation table, the ID.

This PR adds a new attribute to the form model to store the organisation ID. To fully migrate will require other changes in other PRs, once the data has been copied across successfully we can then remove the old `org` attribute that held the slug.

Trello card: https://trello.com/c/pt9lphfe

### DIFF
--- a/spec/request/api/v1/forms_controller_spec.rb
+++ b/spec/request/api/v1/forms_controller_spec.rb
@@ -393,4 +393,57 @@ describe Api::V1::FormsController, type: :request do
       end
     end
   end
+
+  describe "#update_organisation_for_creator" do
+    let(:selected_creator_id) { 1234 }
+    let(:updated_organisation_id) { 111 }
+
+    before do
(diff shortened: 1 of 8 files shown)


spec/request/api/v1/forms_controller_spec.rb
jamie-o-wilkinson, line 1: Should we add a test for querying by `organisation_id`?

lfdebrux: Good idea!

lfdebrux: I've updated the commit to add those tests

jamie-o-wilkinson (approved):
Looks good to me, nice one! 🥇

PR #298, opened 2023-07-10
Add URL query params to logging messages

lfdebrux:

When a HTTP request includes query parameters it is useful to be able to see these in our logs. By default Lograge filters this out as it can contain sensitive information [[1]], but we don't have anything like that in our system.

This is especially useful at the moment when we are changing from using `org` slug to `organisation_id` (see PR #295, ticket https://trello.com/c/pt9lphfe), as without this logging we can't see what the admin app is sending to the api app for requests to `/forms`.

[1]: https://github.com/roidrage/lograge/blob/09bd96e1cb56f94841a64473e4195d413ce908fc/lib/lograge/log_subscribers/action_controller.rb#L21-L39

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -89,6 +89,7 @@ Rails.application.configure do
       h[:request_id] = event.payload[:request_id]
       h[:requested_by] = event.payload[:requested_by] if event.payload[:requested_by]
       h[:form_id] = event.payload[:form_id] if event.payload[:form_id]
+      h[:params] = event.payload[:params].except(:controller, :action)
     end
   end
 


config/environments/production.rb
```diff
@@ -89,6 +89,7 @@
       h[:request_id] = event.payload[:request_id]
       h[:requested_by] = event.payload[:requested_by] if event.payload[:requested_by]
       h[:form_id] = event.payload[:form_id] if event.payload[:form_id]
+      h[:query] = event.payload[:path].split("?").second
```
danworth, line 92: This will work but parsing the params into a json object/hash of key values will make using them in Splunk in future simpler (I'd hope that the logging formatter accepts objects for the values and converts appropriately but I cannot be sure without testing it). I have it in mind to tweak our Splunk set up so that by default it parses the json format of the app logs into json and displays it neatly (this makes searching and extracting values much simpler), so it'll look like this (I've added in the query params just to demonstrate:

image

Then it'll be simple to query such as:
```
| spath output=organisation_id query.organisation_id
```

If its kept as a string then we'd need to `rex` it out, something like
```
rex field=_raw "organisation_id=(?<organisation_id>.*?)[&$]"
```

danworth: Just looking at the Lograge docs they mention that the event payload includes `params` which is the hash of params (its in the what it doesn't do section but in the sense of, if you want it to then you can do it yourself): https://github.com/roidrage/lograge#what-it-doesnt-do

>The payload does already contain the params hash, so you can easily add it in manually using custom_options:
```
params: event.payload[:params].except(*exceptions)
```

Might be helpful?

lfdebrux: Oh nice, I did see this in the docs but wasn't sure if Splunk would handle the nested JSON okay. I like this a lot, I'll do this.

lfdebrux: I've updated the commit to do this 👍

danworth: Great stuff, I think its worth deploying this into dev and we can confirm the log format all the way through into Splunk. I'm happy to help check it out once there are some logs?

danworth: Looks good to me
```
{ [-]
   action: index
   controller: Api::V1::FormsController
   db: 1850.7
   duration: 2713.86
   format: json
   host: api.dev.forms.service.gov.uk
   method: GET
   params: { [-]
     org: government-digital-service
   }
   path: /api/v1/forms
   request_id: 56623f7b-c6ec-4a4a-b823-a987ddcaf57a
   requested_by: forms-admin-133
   status: 200
   view: 0.17
}
```

also works without params in request

```
{ [-]
   action: ping
   controller: HeartbeatController
   db: 0
   duration: 0.33
   format: html
   host: 10.10.5.122
   method: GET
   params: { [-]
   }
   path: /ping
   request_id: fcf62340-8188-434a-bfbc-1fbad8132c96
   status: 200
   view: 0.07
}
```

danworth (approved):
Looks good and the json formatting will help us in Splunk to make simple use of the data in future. The risk is if the params contained sensitive or secret data which is why they're not included by default but you've mentioned doing due diligence to confirm this isn't the case for us with forms-api at least. 

This will help with our observability, nice change 👍

PR #299, opened 2023-07-17
Refactor build files

chao-xian:

Bringing the codebase closer to Rails convention with how it runs tasks in setting up and running tests.

Trello card:

None

#### Checklist

  - [x] README.md
  - [ ] Elsewhere (please link)


chao-xian:
I realise this makes it out of step with the other repos but I’ll raise PRs for those too.

.rspec
aliuk2012, line 1: why is this being deleted?

chao-xian: The document format isn't that useful as a default as I doubt most people read it all when tests are passing. Any failing specs will output the documentation anyway. It's just noisy.

chao-xian: (We don't tend to output the document format on govuk for example)

aliuk2012: Makes sense, thanks for the explanation.  👍

aliuk2012 (approved):

chao-xian:
@aliuk2012 thanks Al! I'll raise corresponding PRs for the other repos when I get a chance

jamie-o-wilkinson (approved):

DavidBiddle (approved):

PR #300, opened 2023-07-19
Structured Logging

danworth:
Structured logging allows simpler searching and reporting from our logs. Introduces a JSON log formatter as per the implementation in forms-runner.

Similar implementation as per: https://github.com/alphagov/forms-runner/pull/371

We should explore ways of sharing common code, happy to revisit this after adding to forms-admin to make sure we have the correct level of reuse before committing to such a thing.

### How to Test 
This is running in development environment and produces logs:
image

- You can run this locally and see the logs in JSON format.
- This is currently deployed in dev environment. 

#### Checklist
Trello card: https://trello.com/c/VSdLnRsI/130-firebreak-convert-logs-to-json-structured

  - [x] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/lib/json_log_formatter.rb
+++ b/app/lib/json_log_formatter.rb
@@ -1,0 +1,25 @@
+require "active_support/logger"
+
+class JsonLogFormatter < ActiveSupport::Logger::Formatter
+  def call(severity, timestamp, _progname, message)
+    log_event = {
+      level: severity,
+      time: timestamp,
+    }
+
+    if is_json(message)
+      log_event.merge!(JSON.parse(message))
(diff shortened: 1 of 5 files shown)


app/lib/json_log_formatter.rb
lfdebrux, line 1: It's annoying that we can't reuse this code across our apps more easily :(

danworth: Yeah my plan is to add this to forms-admin then once we have a complete picture of what can/should be reused to explore what we can do about it 👍 Thanks for the review

PR #316, opened 2023-08-22
Remove 'additional' from guidance_markdown column

SamJamCul:
Renaming the `additional_guidance_markdown` column to `guidance_markdown`. There's a corresponding change to admin here https://github.com/alphagov/forms-admin/pull/602

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -58,10 +58,10 @@ class Page < ApplicationRecord
 private
 
   def additional_guidance_fields
-    if page_heading.present? && additional_guidance_markdown.blank?
-      errors.add(:additional_guidance_markdown, "must be present when Page Heading is present")
-    elsif additional_guidance_markdown.present? && page_heading.blank?
-      errors.add(:page_heading, "must be present when Additional Guidance Markdown is present")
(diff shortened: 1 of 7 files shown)


app/models/page.rb
```diff
@@ -58,9 +58,9 @@ def has_routing_errors
 private
 
   def additional_guidance_fields
-    if page_heading.present? && additional_guidance_markdown.blank?
-      errors.add(:additional_guidance_markdown, "must be present when Page Heading is present")
-    elsif additional_guidance_markdown.present? && page_heading.blank?
+    if page_heading.present? && guidance_markdown.blank?
+      errors.add(:guidance_markdown, "must be present when Page Heading is present")
+    elsif guidance_markdown.present? && page_heading.blank?
       errors.add(:page_heading, "must be present when Additional Guidance Markdown is present")
```
aliuk2012, line 64:
```suggestion
      errors.add(:page_heading, "must be present when Guidance Markdown is present")
```

SamJamCul: Error message has been updated

aliuk2012 (approved):

PR #321, opened 2023-08-30
Rename additional_guidance* vars etc to guidance

aliuk2012:
We should no longer be referring to "additional guidance" in our codebase and instead should be referring to it as just guidance

Trello card: 


jamie-o-wilkinson (approved):
👍

SamJamCul (approved):

PR #322, opened 2023-08-30
Add validation for guidance_markdown

aliuk2012:

We are pulling in `govuk-forms-markdown` gem for validations. Forms-admin will use the same validations that are in the gem but this is really a backstop to ensure that no invalid guidance_markdown is stored in the db and if it is invalid it should flag it up as an error.

Trello card: https://trello.com/c/CCmWlsUJ/958-implementing-markdown-in-forms-api

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ WORKDIR /app
 RUN apk update
 RUN apk upgrade --available
 RUN apk add libc6-compat openssl-dev build-base libpq-dev
+# Adding git so that we can bundle install gems from github
+RUN apk add git
 RUN adduser -D ruby
 
 USER ruby
--- a/spec/models/page_spec.rb
+++ b/spec/models/page_spec.rb
@@ -61,6 +61,31 @@ RSpec.describe Page, type: :model do
         expect(page).to be_invalid
         expect(page.errors[:page_heading]).to include("must be present when Guidance Markdown is present")
       end
+
+      context "when markdown is too long" do
+        it "adds an error to guidance_markdown" do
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -5,6 +5,8 @@ WORKDIR /app
 RUN apk update
 RUN apk upgrade --available
 RUN apk add libc6-compat openssl-dev build-base libpq-dev
+# Adding git so that be can bundle install gems from github
```
DavidBiddle, line 8:
```suggestion
# Adding git so that we can bundle install gems from github
```

spec/models/page_spec.rb
```diff
@@ -61,6 +61,31 @@
         expect(page).to be_invalid
         expect(page.errors[:page_heading]).to include("must be present when Guidance Markdown is present")
       end
+
+      context "when markdown is too long" do
+        it "adds an error to guidance_markdown" do
+          page.guidance_markdown = "ABC" * 5000
+          expect(page).to be_invalid
+          expect(page.errors[:guidance_markdown]).to include("is too long (maximum is 4999 characters)")
+        end
```
SamJamCul, line 70: It's tempting to suggest that the length limit should be something that's defined within the gem, and we can draw from it if we want to pick out the limit - might make a card for it if we end up having to change the limit a bunch

SamJamCul (approved):

spec/models/page_spec.rb
aliuk2012: Yeah, I agree. Lets keep it simple for the moment until we do other validation/customisations

PR #332, opened 2023-09-18
Add make live validation

jamie-o-wilkinson:
Currently we have validation in forms-admin that prevents making a form live until certain conditions are met. However the api will happily make a form live if those conditions aren’t met.

This PR adds an updated task status service from forms-admin, and uses it validate the status of mandatory tasks prior to making a form live. It also adds task status data to the form data itself.

Unfortunately the submission email status depends on data in the forms-admin db, so this isn't included in the validation. Instead the submission email check is still done in forms-admin, and is combined with the `incomplete_tasks` and `task_statuses` from forms-api.

It would be great to hear other people's opinions on whether this refactor (in effect splitting the validation across forms-api and forms-admin) is preferable to duplicating the validation and removing the need to pass data back via the form object.

Trello card: https://trello.com/c/xEyusUhB

### DIFF
--- a/app/controllers/api/v1/forms_controller.rb
+++ b/app/controllers/api/v1/forms_controller.rb
@@ -37,8 +37,12 @@ class Api::V1::FormsController < ApplicationController
   end
 
   def make_live
-    form.make_live!
-    render json: { success: true }.to_json, status: :ok
+    if form.ready_for_live
+      form.make_live!
+      render json: { success: true }.to_json, status: :ok
+    else
+      render json: form.missing_sections.to_json, status: :forbidden
+    end
   end
 
   def show_live
--- a/app/controllers/api/v1/forms_controller.rb
+++ b/app/controllers/api/v1/forms_controller.rb
@@ -41,7 +41,7 @@ class Api::V1::FormsController < ApplicationController
       form.make_live!
       render json: { success: true }.to_json, status: :ok
     else
-      render json: form.missing_sections.to_json, status: :forbidden
+      render json: form.incomplete_tasks.to_json, status: :forbidden
     end
   end
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -52,7 +52,7 @@ class Form < ApplicationRecord
   end
 
   def as_json(options = {})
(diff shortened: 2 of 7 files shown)


app/models/form.rb
```diff
@@ -52,7 +52,7 @@ def name=(val)
   end
 
   def as_json(options = {})
-    options[:methods] ||= %i[live_at start_page has_draft_version has_live_version has_routing_errors]
+    options[:methods] ||= %i[live_at start_page has_draft_version has_live_version has_routing_errors ready_for_live incomplete_tasks task_statuses]
```
jamie-o-wilkinson, line 55: I'm not sure I like this, it feels like it might pollute the form object computed data too much...

app/controllers/api/v1/forms_controller.rb
jamie-o-wilkinson, line 1: Unfortunately we can't gate making live on the submission email status, as the state of the email and confirmation is stored in the forms-admin db. This somewhat lessens the value of this whole refactor in my opinion, but I would like to hear what others think.

app/service/task_status_service.rb
```diff
@@ -0,0 +1,85 @@
...
+      pages_status:,
+      declaration_status:,
+      what_happens_next_status:,
+      privacy_policy_status:,
+      support_contact_details_status:,
+      make_live_status:,
+    }
+  end
+
+private
```
jamie-o-wilkinson, line 29: None of the statuses actually need to be accessed directly any more so can be made private

app/models/form.rb
thomasiles: does it make sense because the API is mainly about **editing** forms rather than live forms?

I guess live forms don't need this info and the options are overridden to not include it in those?

thomasiles (approved):
I think this makes things a lot simpler and it reads well. Nice one!

app/models/form.rb
jamie-o-wilkinson: When you put it like that, yes I suppose the sole purpose of this is to serve forms-admin what it needs. I just don't want to see an ever expanding list of dynamic attributes that get passed back as JSON. 😅 

Yeah the added dynamic attributes aren't included in the JSON snapshot made when forms are made lives 👍

PR #336, opened 2023-10-03
Change API endpoints to respond with resource

lfdebrux:

Trello card: https://trello.com/c/1Wr1DCBT 

 We want the API to respond with a resource when requests are made, even if it's a request to update or delete an object. As well as being more in line with other APIs (see the JSON::API spec for example [[1]]), this will also let us reduce the number of requests we need to make in the admin app.

For the simple traditional CRUD (Create, Read, Update, and Delete) operations the conventional responses are prescribed in the JSON::API spec. For some of our endpoints, they don't really fit into the RESTful/CRUD model, so I've just gone with what I think is most predictable/useful.

Note that technically the changes in this PR are breaking changes and should result in a new API version, but currently our API is internal only and none of the consumers rely on the response body data, so no breakage has occured and I think we can get away without the version bump.

[1]: https://jsonapi.org/format/

I've tested these changes locally and in dev using the end to end tests.

### DIFF
--- a/spec/request/api/v1/conditions_controller_spec.rb
+++ b/spec/request/api/v1/conditions_controller_spec.rb
@@ -52,7 +52,7 @@ describe Api::V1::ConditionsController, type: :request do
     it "returns condition id, status code 201 when new condition created" do
       expect(response.status).to eq(201)
       expect(response.headers["Content-Type"]).to eq("application/json")
-      expect(json_body).to eq({ id: new_condition.id })
+      expect(json_body).to include(id: new_condition.id, **new_condition_params)
     end
 
     context "with a form with question_section_completed = true" do
@@ -116,7 +116,7 @@ describe Api::V1::ConditionsController, type: :request do
     it "returns correct response" do
       expect(response.status).to eq(200)
       expect(response.headers["Content-Type"]).to eq("application/json")
-      expect(json_body).to eq({ success: true })
+      expect(json_body).to include(**params)
       expect(condition.reload.answer_value).to eq(answer_value)
       expect(form.reload.question_section_completed).to be false
     end
@@ -128,9 +128,7 @@ describe Api::V1::ConditionsController, type: :request do
     end
 
     it "removes the condition and returns the correct response" do
-      expect(response.status).to eq(200)
-      expect(response.headers["Content-Type"]).to eq("application/json")
-      expect(json_body).to eq({ success: true })
+      expect(response.status).to eq(204)
       expect(routing_page.routing_conditions.count).to eq(0)
       expect(form.reload.question_section_completed).to be false
     end
--- a/spec/request/api/v1/conditions_controller_spec.rb
+++ b/spec/request/api/v1/conditions_controller_spec.rb
@@ -50,7 +50,7 @@ describe Api::V1::ConditionsController, type: :request do
     end
 
     it "returns condition id, status code 201 when new condition created" do
-      expect(response.status).to eq(201)
+      expect(response).to have_http_status(:created)
       expect(response.headers["Content-Type"]).to eq("application/json")
       expect(json_body).to include(id: new_condition.id, **new_condition_params)
     end
@@ -114,7 +114,7 @@ describe Api::V1::ConditionsController, type: :request do
     end
 
     it "returns correct response" do
-      expect(response.status).to eq(200)
+      expect(response).to have_http_status(:ok)
       expect(response.headers["Content-Type"]).to eq("application/json")
       expect(json_body).to include(**params)
       expect(condition.reload.answer_value).to eq(answer_value)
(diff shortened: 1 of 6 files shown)


spec/request/api/v1/conditions_controller_spec.rb
```diff
@@ -116,7 +116,7 @@
     it "returns correct response" do
       expect(response.status).to eq(200)
```
aliuk2012, line 117:
```suggestion
      expect(response).to have_http_status(:ok)
```

spec/request/api/v1/conditions_controller_spec.rb
```diff
@@ -52,7 +52,7 @@
     it "returns condition id, status code 201 when new condition created" do
       expect(response.status).to eq(201)
```
aliuk2012, line 53:
```suggestion
      expect(response).to have_http_status(:created)
```

spec/request/api/v1/forms_controller_spec.rb
```diff
@@ -125,7 +125,7 @@
       it "returns a status code 201" do
         expect(response.status).to eq(201)
```
aliuk2012, line 126:
```suggestion
        expect(response).to have_http_status(:created)
```

spec/request/api/v1/forms_controller_spec.rb
```diff
@@ -213,7 +213,7 @@
       put form_path(form1), params: { submission_email: "test@example.gov.uk" }, as: :json
       expect(response.status).to eq(200)
```
aliuk2012, line 214:
```suggestion
      expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -60,7 +60,7 @@
     it "returns page id, status code 201 when new page created" do
       expect(response.status).to eq(201)
```
aliuk2012, line 61:
```suggestion
      expect(response).to have_http_status(:created)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -159,7 +159,7 @@
     it "returns correct response" do
       expect(response.status).to eq(200)
```
aliuk2012, line 160:
```suggestion
      expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -187,7 +187,7 @@
         it "returns correct response" do
           expect(response.status).to eq(200)
```
aliuk2012, line 188:
```suggestion
          expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -208,9 +208,9 @@
       it "returns correct response" do
         expect(response.status).to eq(200)
```
aliuk2012, line 209:
```suggestion
        expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -229,9 +229,7 @@
 
     context "with existing page" do
       it "removes page and returns correct response" do
-        expect(response.status).to eq(200)
-        expect(response.headers["Content-Type"]).to eq("application/json")
-        expect(json_body).to eq({ success: true })
+        expect(response.status).to eq(204)
```
aliuk2012, line 232:
```suggestion
        expect(response).to have_http_status(:no_content)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -278,7 +276,7 @@
       it "returns correct response" do
         expect(response.status).to eq(200)
```
aliuk2012, line 277:
```suggestion
        expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -297,7 +295,7 @@
...
```
aliuk2012, line 296:
```suggestion
        expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -329,7 +327,7 @@
...
```
aliuk2012, line 328:
```suggestion
        expect(response).to have_http_status(:ok)
```

spec/request/api/v1/pages_controller_spec.rb
```diff
@@ -348,7 +346,7 @@
...
```
aliuk2012, line 347:
```suggestion
        expect(response).to have_http_status(:ok)
```

aliuk2012 (commented):
Only a couple suggestions to improve the status code specs

aliuk2012 (approved):
Nice work, 

This does unblock us from some features we wanted to do but the previous implementation meant we would have to make extra requests that could be unreliable

PR #339, opened 2023-10-06
Log page ID with requests and improve logging of form ID

lfdebrux:

When searching Splunk for requests related to a particular page of a form it would be nice to be able to construct a query with `page_id=n`.

This PR adds the page ID to the log event JSON payload so that this will be possible in future.

See https://github.com/alphagov/forms-runner/pull/431 and https://github.com/alphagov/forms-admin/pull/668 for similar PRs for the runner and admin apps.

While adding this functionality I also noticed that we're not consistently logging `form_id` with every request, because the name of the param changes depending on the route. This PR also tries to fix that, so we get the `form_id` key in all relevant requests.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,6 +30,7 @@ class ApplicationController < ActionController::API
                         elsif params[:form_id].present?
                           params[:form_id]
                         end
+    payload[:page_id] = params[:page_id] if params[:page_id].present?
   end
 
 private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,7 +25,11 @@ class ApplicationController < ActionController::API
     payload[:host] = request.host
     payload[:request_id] = request.request_id
     payload[:requested_by] = "#{@access_token.owner}-#{@access_token.id}" if @access_token.present?
-    payload[:form_id] = params[:form_id] if params[:form_id].present?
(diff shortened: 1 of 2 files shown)


chao-xian (approved):

app/controllers/application_controller.rb
```diff
@@ -25,7 +25,12 @@ def append_info_to_payload(payload)
     payload[:host] = request.host
     payload[:request_id] = request.request_id
     payload[:requested_by] = "#{@access_token.owner}-#{@access_token.id}" if @access_token.present?
-    payload[:form_id] = params[:form_id] if params[:form_id].present?
+    payload[:form_id] = if params[:id].present?
```
thomasiles, line 28: I'm not sure we can use :id like this in the application controller. `id` is the name of the param in the conditions controller and the access_tokens controller.

I guess another approach might be to overide `append_info_to_payload` in the forms controller, calling super and setting the correct form_id value. Not 100% sure that would work.

lfdebrux: > I'm not sure we can use :id like this in the application controller. id is the name of the param in the conditions controller and the access_tokens controller.

Ah, good catch.

> I guess another approach might be to overide append_info_to_payload in the forms controller, calling super and setting the correct form_id value. Not 100% sure that would work.

I like this idea a lot, I'm going to give it a try.

PR #346, opened 2023-10-17
Fix logging of form_id in FormsController

lfdebrux:

For requests that are routed to the FormsController, the form ID is named `:id` rather than `:form_id`, so the logging setup needs to be slightly different. This commit adds a specialisation of the `append_info_to_payload` method to the `FormsController` class so we get the correct behaviour when necessary.

Thanks @thomasiles for pointing out the issue and suggesting that we can use OOP to fix this (see https://github.com/alphagov/forms-api/pull/339#discussion_r1358180399).


jamie-o-wilkinson (approved):
Looks good to me 👍

PR #347, opened 2023-10-17
Update various dev dependency gems

aliuk2012:

See individual commits

Trello card: 


DavidBiddle (approved):
Change looks good to me. But out of interest, do you know why Dependabot hasn't handled these updates?

aliuk2012:
@DavidBiddle we setup dependabot to only track prod dependencies.  The reason for this is that dev dependencies are more likely to be bumped more frequently but with very little benefit to us. its better to just bump the dev dependence periodically. Might be worth checking our js dependencies.

PR #349, opened 2023-10-19
Rename param for access tokens route

lfdebrux:

Renames the name of the placeholder in the routes for access tokens from `:id` to `:token_id`, to make it harder to confuse the parameter with the form ID of other routes.

I also expanded the responses from the access token endpoints so that you actually get the ID when you create a token.

### DIFF
--- a/app/controllers/api/v1/access_tokens_controller.rb
+++ b/app/controllers/api/v1/access_tokens_controller.rb
@@ -6,9 +6,9 @@ class Api::V1::AccessTokensController < ApplicationController
 
   def create
     @access_token = AccessToken.new(token_params)
-    users_token = @access_token.generate_token
+    token = @access_token.generate_token
     if @access_token.save
-      render json: { token: users_token }.to_json, status: :created
+      render json: { **@access_token.as_json, token: }.to_json, status: :created
     else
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/access_tokens_controller.rb
```diff
@@ -35,6 +35,6 @@ def token_params
   end
 
   def token_deactivate_params
-    params.require(:id)
+    params.require(:token)
```
aliuk2012, line 38: The param is being used to find `AccessToken` by id. This is not the same as a form id. I know all the specs are passing but have you tried this locally to see if it works?

lfdebrux: I see, I was confused by the fact that this is the only place we refer to the `AccessToken` ID in our code (it's not in the model, or the schema file, but I realise now it's just implied). Thanks for pointing this out, I'll redo this PR.

lfdebrux: I've changed the rename to be to `token_id`, I've also tweaked the API a bit to be more conventional. I've tested all this in dev.

DavidBiddle (approved):
Seems like a sensible change and the implementation looks good to me! 🎉

PR #361, opened 2023-11-10
Add database seed with all question types form

lfdebrux:

Make it quicker to get a useful test form created when doing local development.

This PR adds a copy of @DavidBiddle's 'All question types' form that already exists in dev at `/forms/71/live`. If you want to add this to your database you can simply run `rails db:seed`

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,70 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+
+submission_email = ENV["EMAIL"] || `git config --get user.email`.strip
+organisation_id = 1 # Assumes you're using the default database seed for forms-admin
+
+all_question_types_form = Form.create!(
+  name: "All question types form",
+  pages: [
+    Page.create(
+      question_text: "Single line of text",
+      answer_type: "text",
+      answer_settings: {
+        input_type: "single_line",
+      },
+    ),
+    Page.create(
+      question_text: "Number",
+      answer_type: "number",
+    ),
+    Page.create(
+      question_text: "Address",
+      answer_type: "address",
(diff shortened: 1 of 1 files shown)


chao-xian (dismissed):
TYSM for doing this!!!! Love it!

DavidBiddle (approved):
This looks good, and a very nice feature to have - only thing worth noting is that it doesn't include a selection question, so 'all question types form' is a bit of a misnomer (which is a fault with the original form rather than with this PR)

lfdebrux:
> This looks good, and a very nice feature to have - only thing worth noting is that it doesn't include a selection question, so 'all question types form' is a bit of a misnomer (which is a fault with the original form rather than with this PR)

Oh good catch. I'll add a selection question now.

db/seeds.rb
```diff
@@ -5,3 +5,83 @@
...
+    ),
+    Page.create(
+      question_text: "Phone number",
+      answer_type: "phone_number",
+    ),
+    Page.create(
+      question_text: "Selection from a list of options",
+      answer_type: "selection",
+      answer_settings: {
+        "only_one_option": "0", # TODO: investigate why we set this to "0"
```
lfdebrux, line 59: I've created a ticket for this: https://trello.com/c/prWyiTVd/1310-forms-admin-saves-onlyoneoption-0-into-database-if-user-does-not-check-box

lfdebrux:
@DavidBiddle I've added a selection question, while doing so I found a curiosity which I have recorded on Trello. Would you mind reviewing, or can I get a re-review from @chao-xian?

db/seeds.rb
DavidBiddle: I think I probably introduced this - I vaguely remember running into trouble with these boolean fields and ActiveResource. There's almost certainly a better way of doing it, so I appreciate you adding the card to investigate it!

DavidBiddle (approved):

PR #362, opened 2023-11-10
Update make_live_form spec descriptions

chao-xian:
This was confusing before, and we are about to add a spec on making forms un-live without deleting the parent form.

### DIFF
--- a/spec/models/form_spec.rb
+++ b/spec/models/form_spec.rb
@@ -14,12 +14,14 @@ RSpec.describe Form, type: :model do
     end
   end
 
-  describe "made_live_form" do
-    let(:made_live_form) { create :made_live_form }
-    let(:form) { made_live_form.form }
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):

spec/models/form_spec.rb
```diff
@@ -14,12 +14,14 @@
     end
   end
 
-  describe "made_live_form" do
-    let(:made_live_form) { create :made_live_form }
-    let(:form) { made_live_form.form }
+  describe ".destroy" do
+    context "when a form was made live" do
```
lfdebrux, line 18: What's the value of the extra `context` here?

chao-xian: Because I changed the `describe` of the spec as it's about when the destroy method is called, adding the context that the state of the form is one that has a `made_live_form` child.

PR #363, opened 2023-11-10
Add code that makes forms un-live

chao-xian:

Trello card: https://trello.com/c/E6iLOdFK/1180-add-a-un-alive-switch-for-a-form

If a form hits any sort of trouble, we need a reliable way to make it offline

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -87,6 +87,10 @@ class Form < ApplicationRecord
 
   delegate :task_statuses, to: :task_status_service
 
+  def make_unlive!
+    made_live_forms.destroy_all
+  end
+
 private
 
   def task_status_service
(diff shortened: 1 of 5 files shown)


app/models/form.rb
lfdebrux, line 1: Is the `has_live_version` attribute used anywhere in forms-admin?

chao-xian: [ARGH](https://github.com/search?q=repo%3Aalphagov%2Fforms-admin%20has_live_version&type=code)

chao-xian: I'll drop that commit and do it later...

lfdebrux (approved):
Tested locally, works for me 👍

PR #364, opened 2023-11-13
Add rake task to make a form not live

chao-xian:

This is a follow on from https://github.com/alphagov/forms-api/pull/363 and adds a rake task to call that code to make a form no longer live.

Trello card: https://trello.com/c/E6iLOdFK/1180-add-a-un-alive-switch-for-a-form

### DIFF
--- a/lib/tasks/forms_admin.rake
+++ b/lib/tasks/forms_admin.rake
@@ -1,0 +1,16 @@
+namespace :forms_admin do
+  desc "Unlive a form"
+  task :make_unlive, [:form_id] => :environment do |_, args|
+    form = Form.find(args[:form_id])
+
+    if form.made_live_forms.empty?
+      puts "Form #{form.name} is not live" unless Rails.env.test?
+      exit
+    end
+
+    form.make_unlive!
+    form.reload
+
+    puts "Unlived form #{form.name}" unless Rails.env.test?
+  end
+end
--- a/spec/lib/tasks/forms_admin_spec.rb
+++ b/spec/lib/tasks/forms_admin_spec.rb
@@ -1,0 +1,31 @@
+require "rails_helper"
+require "rake"
+
+RSpec.describe "forms_admin.rake" do
+  describe "forms_admin:make_unlive", type: :task do
+    before do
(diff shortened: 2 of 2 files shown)


lib/tasks/forms_admin.rake
```diff
@@ -0,0 +1,14 @@
+namespace :forms_admin do
+  desc "Unlive a form"
+  task :make_unlive, [:form_id] => :environment do |_, args|
+    form = Form.find(args[:form_id])
+    form.make_unlive!
+    form.reload
+
+    if form.has_live_version
```
DavidBiddle, line 8: I think I'd expect `form.has_live_version` to always return `false` if we've already run `form.make_unlive!` in the task - if it doesn't then I feel like it might not be checking the right thing

DavidBiddle:
```ruby
def has_live_version
  made_live_forms.present?
end
```
I think that'll return `true` if `made_live_forms` is `[]`, which I think might be what `made_live_forms` returns when we've run `made_live_forms.destroy_all`

DavidBiddle: So the `else` statement here won't get much action, and I think that would also explain why you're not seeing `Form has no live version` when running `make_unlive` on a form that's already been un-alived

chao-xian: Hmmm but `.present?` should do:
```ruby
irb(main):001> [].present?
=> false
```

DavidBiddle: Ah, you're right. I'm now wondering what made_live_forms is returning then!

chao-xian: Ahhhhhhh!!!!  😖 
```ruby
irb(main):001> form = Form.find(20)
irb(main):004> form.has_live_version
=> false
```

chao-xian: It took me far to long to see that the logic was in the wrong place 🥲

DavidBiddle (dismissed):
Looks good! You might want to add a test for the case where the form isn't live, but I don't think it's a blocker

DavidBiddle (approved):

PR #369, opened 2023-11-14
Update rails 7.1.2

chao-xian:

Trello card: 


config/environments/production.rb
```diff
@@ -38,8 +37,25 @@
   # config.action_cable.url = "wss://example.com/cable"
   # config.action_cable.allowed_request_origins = [ "http://example.com", /http:\/\/example.*/ ]
 
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
```
aliuk2012, line 42:
```suggestion
  config.assume_ssl = true
```

aliuk2012:
@chao-xian Why are there DB migrations in this PR? We have never used ActiveStorage so wondering why we need them.

chao-xian:
@aliuk2012 according to the `Gemfile.lock` it's a dependency of `actionmailbox` and `actiontext` and `rails` itself it seems. The upgrade task added these files.

aliuk2012:
@chao-xian  🤔  I mean we don't specifically have a table called `active_storage_blobs` so I suppose those migrations aren't going to run anyway because there a bunch of `return unless table_exists?` statements

chao-xian:
@aliuk2012 oh! Good point. Lemme drop them...

config/environments/production.rb
```diff
@@ -38,8 +37,25 @@
...
-  # config.force_ssl = true
+  config.force_ssl = true
+
+  # Log to STDOUT by default
+  config.logger = ActiveSupport::Logger.new($stdout)
+    .tap  { |logger| logger.formatter = ::Logger::Formatter.new }
+    .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
+
+  # Prepend all log lines with the following tags.
+  config.log_tags = [:request_id]
```
danworth, line 53: This is still draft but just in case its helpful to point out, there is some default logging configuration in config/application.rb and from what I remember Rails' `log_tags` feature does not work with lograge and structured/json logging because it mangles the formatting by prepending the tags outside of the provided formatter like `[rails log tags] {then the log message}`. We've configured lograge to include the request_id and other useful things so log_tags shouldn't be necessary (unless we overhaul things a bit): https://github.com/alphagov/forms-api/blob/main/app/controllers/application_controller.rb#L23

chao-xian: I wasn't aware of that. I'll remove this incoming change from the Rails upgrade.

spec/spec_helper.rb
```diff
@@ -1,3 +1,4 @@
+require "active_support"
 require "active_support/testing/time_helpers"
```
aliuk2012, line 2: 🤔  if we pulling in `active_support` above we might not need to explicitly require this?

chao-xian: it completely broke without it I've no idea why it has to be so explicit so keeping it in

chao-xian:
Deployed to dev and ran e2e tests which passed.

config/environments/production.rb
```diff
@@ -38,8 +37,25 @@
...
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  config.assume_ssl = true
+
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
+
+  # Log to STDOUT by default
+  config.logger = ActiveSupport::Logger.new($stdout)
```
jamie-o-wilkinson, line 48: How does this play with config/application.rb:54-57? It looks like prior to the upgrade we checked the `RAILS_LOG_TO_STDOUT` env var before defaulting to using stdout

chao-xian: https://github.com/alphagov/forms-deploy/pull/395

ah I didn't take it out of https://github.com/alphagov/forms-product-page/blame/main/config/application.rb#L51 which I'll do

danworth: This seems related to the PR I have open (and just want to check in Dev once more before merging) https://github.com/alphagov/forms-product-page/pull/139

chao-xian: We don't have specs for the lograge stuff yet but I'll do that as a separate PR instead of part of the upgrade

DavidBiddle (approved):
Code looks good to me and no issues locally 🎉

lfdebrux:
When I run the tests locally for this branch I get a bunch of app log output in the terminal alongside the RSpec output. I can see the same thing in CI: https://github.com/alphagov/forms-api/actions/runs/7221736387/job/19677343512?pr=369.

chao-xian:
@lfdebrux as per our chat and have also addressed in 
* https://github.com/alphagov/forms-product-page/pull/169/
* https://github.com/alphagov/forms-runner/pull/552/

aliuk2012 (approved):

PR #370, opened 2023-11-15
Remove timestamp constraint

chao-xian:
These were only made explicit to support a potential rollback to Grape as we migrated from there to Rails in the past. However this is now causing us issues as part of our Rails 7.1 upgrade [1]

[1]: https://github.com/alphagov/forms-api/actions/runs/6876807358/job/18703161928?pr=369

### DIFF
--- a/db/migrate/20221219153156_create_pages.rb
+++ b/db/migrate/20221219153156_create_pages.rb
@@ -1,6 +1,6 @@
 class CreatePages < ActiveRecord::Migration[7.0]
   def change
-    create_table :pages, if_not_exists: true do |t|
+    create_table :pages do |t|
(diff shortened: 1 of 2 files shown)


db/migrate/20221219153156_create_pages.rb
```diff
@@ -13,6 +13,5 @@ def change
     end
 
     add_reference :pages, :form, foreign_key: true, if_not_exists: true
```
aliuk2012, line 15:
```suggestion
    add_reference :pages, :form, foreign_key: true
```

chao-xian: damn! thanks for spotting

PR #371, opened 2023-11-16
Add empty what happens next markdown field

DavidBiddle:

Trello card: https://trello.com/c/AbPj2LRr/1191-allow-markdown-on-confirmation-page-what-happens-next-content

Adds a null `what_happens_next_markdown` field to the form. Also adds this null field to the most recent `made_live_form`.

To do in other PRs:
- add logic in the runner (https://github.com/alphagov/forms-runner/pull/496) and admin to use the markdown field if it's present, and the HTML if not 
- convert the HTML content in the API into markdown and store it in this new field
- cleanup: delete the old field and all the logic related to it which we no longer need

### DIFF
--- a/db/migrate/20231116142411_add_what_happens_next_markdown_to_forms.rb
+++ b/db/migrate/20231116142411_add_what_happens_next_markdown_to_forms.rb
@@ -1,0 +1,20 @@
+class AddWhatHappensNextMarkdownToForms < ActiveRecord::Migration[7.0]
+  def change
+    add_column :forms, :what_happens_next_markdown, :text
+
+    reversible do |direction|
+      MadeLiveForm.find_each do |made_live_form|
+        form_blob = JSON.parse(made_live_form.json_form_blob, symbolize_names: true)
+
+        direction.up do
+          form_blob[:what_happens_next_markdown] = nil
+        end
+        direction.down do
+          form_blob.delete(:what_happens_next_markdown)
+        end
+
+        made_live_form.update!(json_form_blob: form_blob.to_json)
+      end
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_08_24_153800) do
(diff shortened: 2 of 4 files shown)


db/migrate/20231116142411_add_what_happens_next_markdown_to_forms.rb
```diff
@@ -0,0 +1,5 @@
+class AddWhatHappensNextMarkdownToForms < ActiveRecord::Migration[7.0]
+  def change
+    add_column :forms, :what_happens_next_markdown, :string
```
aliuk2012, line 3:
```suggestion
    add_column :forms, :what_happens_next_markdown,  :text
```

db/migrate/20231116142411_add_what_happens_next_markdown_to_forms.rb
```diff
@@ -0,0 +1,23 @@
+class AddWhatHappensNextMarkdownToForms < ActiveRecord::Migration[7.0]
+  def change
+    add_column :forms, :what_happens_next_markdown, :text
+
+    reversible do |direction|
+      Form.find_each do |form|
+        if form.made_live_forms.present?
```
aliuk2012, line 7: Why do we need to use forms to get to live versions? I think we should just go through all the live versions
pseudo  code changes so we dont just end up with the last 
```suggestion
      MadeLiveForm.find_each do |form|
```

DavidBiddle: @aliuk2012 I copied it from a previous migration (`20230727151012_update_organisation_id_for_made_live_forms.rb`). I've updated it now though

aliuk2012:
Does this have to change? https://github.com/alphagov/forms-api/blob/main/db/seeds.rb#L85

DavidBiddle:
@aliuk2012 I don't think so at the moment - the seed still works and sets `what_happens_next_markdown` to `null` in the database, which is the default value (which we want because that's the state the forms in prod will be in after the migration). Good spot though - I'll add a note to the card so that I remember to update the seed when we do the conversion to markdown.

aliuk2012 (approved):
I've tested this locally and everything looks good. I tried using form-runner from main branch and also the feature branch and both cases worked perfectly.

PR #377, opened 2023-11-28
Use what_happens_next markdown in task completion check

DavidBiddle:

As part of the ready_for_live check we test whether the form user has provided what happens next information. This previously only checked whether the what_happens_next_text field was populated, but we also want it to return true if what_happens_next markdown is populated as well.

Trello card: Part of https://trello.com/c/AbPj2LRr/1191-allow-markdown-on-confirmation-page-what-happens-next-content

### DIFF
--- a/app/service/task_status_service.rb
+++ b/app/service/task_status_service.rb
@@ -53,7 +53,7 @@ private
   end
 
   def what_happens_next_status
-    if @form.what_happens_next_text.present?
+    if @form.what_happens_next_text.present? || @form.what_happens_next_markdown.present?
       :completed
     else
       :not_started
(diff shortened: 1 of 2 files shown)


SamJamCul (approved):
Looks good, could also be worth having a test which checks for the else `:not_started` bit

DavidBiddle:
@SamJamCul just went to add one and it turned out there was already one in there 😁  https://github.com/alphagov/forms-api/blob/eede7e4fc2598feee5e2d7917c74bf912c874f5a/spec/service/task_status_service_spec.rb#L23

SamJamCul:
> @SamJamCul just went to add one and it turned out there was already one in there 😁
> 
> https://github.com/alphagov/forms-api/blob/eede7e4fc2598feee5e2d7917c74bf912c874f5a/spec/service/task_status_service_spec.rb#L23

Ah weird I completely didn't see it, definitely there though!

PR #378, opened 2023-12-01
Convert what happens next content to markdown

DavidBiddle:

Trello card: https://trello.com/c/AbPj2LRr/1191-allow-markdown-on-confirmation-page-what-happens-next-content

Adds a rake task which takes the content of the what_happens_next_text field, converts it to markdown using the ReverseMarkdown gem, then saves this markdown in the to the what_happens_next_markdown field that we added in #371. It also adds this markdown to the made live form.

Doing this as a rake task means we're not stuck with a migration that's dependent on a specific gem, and we've previously said we'd consider reserving migrations for schema changes. The downside is we'll have to run the task manually in each environment.

To do in other PRs:
- cleanup: delete the old field and all the logic related to it which we no longer need

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,3 +81,5 @@ group :test do
   # Code coverage reporter
   gem "simplecov", require: false
 end
+
+gem "reverse_markdown", "~> 2.1"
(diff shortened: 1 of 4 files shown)


aliuk2012 (approved):
Nice work, I like the specs created for the rake task

PR #388, opened 2023-12-15
API Token tweaks

lfdebrux:

While working on #368 I realised that we don't have anything preventing the same API token being used more than once. This seems like a bug to me.

I've also thought it would make API keys easier to identify if they have a prefix, the same way GitHub or Trello does, so I added that as well.

### DIFF
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -5,7 +5,7 @@ class AccessToken < ApplicationRecord
   scope :active, -> { where(deactivated_at: nil) }
 
   def generate_token
-    users_token = SecureRandom.uuid
+    users_token = "forms_#{SecureRandom.uuid}"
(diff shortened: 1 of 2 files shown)


DavidBiddle (approved):
Nice, think both of these changes make sense. I like the fact the test allows for the digest to be the same, I'd never have thought of that!

lfdebrux:
> Hi @lfdebrux Can I double check why this is needed? Have we introduced some way for devs to insert their own generated token? I'm asking because originally the idea was that the model would control the value of the token digest being generated so in theory there should never have been duplicate token digests

Yes, see https://github.com/alphagov/forms-api/pull/368

aliuk2012:
thanks @lfdebrux  I've posted a message about allowing devs to use whatever token they like,

PR #390, opened 2023-12-19
Deny access to methods other than GET for readonly access tokens

lfdebrux:

This PR adds a crude form of access control to the API. Access tokens now have associated permissions; either `:all` or `:readonly`. Access tokens with `:readonly` permissions can only be used for `GET` requests.

Otherwise, nothing else should be changed with using the API. The aim is to make us able to feel more free as developers about using the API for debugging or inspecting production systems.


cadmiumcat:
Archiving repo

PR #391, opened 2023-12-19
Rename authentication_key to auth_key

lfdebrux:

In other apps we call this setting `Settings.forms_api.auth_key`. It would be helpful if this setting had the same name in all apps, because we could set it everywhere with one environment variable.

This commit renames the setting key in forms-api to match what is currently in forms-admin and forms-runner.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,9 +32,9 @@ class ApplicationController < ActionController::API
 private
 
   def authenticate_using_old_env_vars
-    return false if request.headers["X-Api-Token"].blank? || Settings.forms_api.authentication_key.blank?
+    return false if request.headers["X-Api-Token"].blank? || Settings.forms_api.auth_key.blank?
 
-    request.headers["X-Api-Token"] == Settings.forms_api.authentication_key
+    request.headers["X-Api-Token"] == Settings.forms_api.auth_key
   end
 
   def authenticate_using_access_tokens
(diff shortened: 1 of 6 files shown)


aliuk2012:
I think `authentication_key` is better than `auth_key` because its clear that its used for authentication when `auth_key` could be seen as `authorisation key` which is a total different thing. I agree we should be consistent across our apps and its also worth nothing that these changes would need to happen in forms-deploy as well.

lfdebrux:
> I think `authentication_key` is better than `auth_key` because its clear that its used for authentication when `auth_key` could be seen as `authorisation key` which is a total different thing. I agree we should be consistent across our apps and its also worth nothing that these changes would need to happen in forms-deploy as well.

Normally I would be all for making sure the meaning of the name is clear, however as the context here is a setting that we'll normally set by environment variable, I think there is an advantage in this case for keeping the shorter name.

DavidBiddle (approved):
This change makes sense to me. Do we need to make a corresponding change in forms-deploy?

lfdebrux:
> This change makes sense to me. Do we need to make a corresponding change in forms-deploy?

@DavidBiddle thanks for the review! We do need to make a change in forms-deploy, however I think as the infra team are currently refactoring the Terraform we probably need to wait a bit before making that change. So I'm going to put this PR into draft until then.

lfdebrux:
I've opened PR https://github.com/alphagov/forms-deploy/pull/526 to make this PR work.

PR #410, opened 2024-01-10
External id spike

SamJamCul:
[This Spike](https://trello.com/c/BruuEN5Z/1225-spike-change-ids-in-urls-of-forms-to-not-be-the-primary-key) looks at how we would modify forms-api in order to stop exposing our database IDs for form objects, and instead expose them with an `external_id`. This ID is randomly generated when a new form is created, and form lookups are done through this rather than the ID when using the API endpoints.


app/models/form.rb
```diff
@@ -93,6 +95,13 @@ def make_unlive!
 
 private
 
+  def set_external_id
+    loop do
+      self.external_id = SecureRandom.uuid[0, 8]
```
lfdebrux, line 100: Any reason not to use `SecureRandom.hex(8)`?

https://ruby-doc.org/3.2.2/stdlibs/random/Random/Formatter.html#method-i-hex

lfdebrux: Woops, I misread the docs, I meant `SecureRandom.hex(4)` 😁

chao-xian (commented):
Appreciate this is a Draft and only a Spike so my comments can be treated as notes.

PR #411, opened 2024-01-11
Update Development Config

danworth:
This makes it simpler to run locally in development mode using default username, password and host for postgres running locally. The additional config.hosts for forms_api:9292 is necessary when running all applications with docker-compose.

Running our apps locally.


DavidBiddle (approved):
Code looks good, and works on my machine with and without Docker 👍

PR #414, opened 2024-01-16
SPIKE - using a state machine for draft/live form

aliuk2012:
### States

```mermaid
stateDiagram
    Draft --> Live
    draft_live: Draft live
    Live --> draft_live
    draft_live --> Live
    draft_archived: Draft archived
    draft_live --> draft_archived
    Live --> Archived
    Archived --> Live
    Archived --> draft_archived
    draft_archived --> Live
 ```

#### Draft
Initial forms can be deleted as they are not official forms yet. Draft forms can be created/edited by trial/editor users. Previews of draft forms are viewable in `forms-runner`. This is the only state that a form created by a trial user can get to until the  trial users account has been upgraded to an Editor role.

#### Live
Form record is transitioned to this state when a user has completed all the required tasks for a form and decides to make a form live. Live forms cannot be HARD deleted as they are now official forms which could have been used by the public to submit their data to a government department. Previews of live forms are viewable in `forms-runner`.  When a form is made live we take a snapshot of the form, all its pages & all the page routing conditions and store it in `made_live_forms` table as JSON blob format. A form can have multiple "live versions" in `made_live_forms` table but we only ever return the latest version to our users. We keep multiple versions so we could serve up previous versions if need be or if users start a form and a new version is published, we should still be able to allow them to complete the version of form they started.

#### Draft live
Form record is transitioned to this state when a user clicks 
image
from the live form show page. This would be a change from the current method which is that we look at the forms `updated_at` and check if it is after the forms `live_at`.
A new draft of a live form but doesn't create a new record in the database. This state identifies when a form creator has made changes to a live form without affecting that live form. Once the form creator has approval for the changes to go live, the form can transition back to `Live` state.  Previews of draft forms are viewable in `forms-runner`. 

In this state, if a form is `archived`, the transition will go to `archived` and then on to `draft_archived` 

#### Archived
A form creator can only archive a form that is in either Live or Draft Live state. In this state a form filler would receive 404 "Page not Found" Error in forms-runner.  

#### Draft Archived
This is similar to how `Draft live` works except when  in this state a form filler would still receive 404 "Page not Found" Error in forms-runner unless they are previewing a draft version of the form. Once a form creator is ready to make a form live they should be able to transition this back to `Live` state.

[Trello](https://trello.com/c/G0bbEVGK/1447-spike-how-should-we-accurately-model-and-track-the-state-of-a-form)

- Adds a new column to forms table
- Adds aasm gem and after_commit_everywhere gem for activerecord support
- Mock out  a state machine for forms
	- reuse existing guards 		

Trello card: 

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,3 +83,7 @@ group :test do
 end
 
 gem "reverse_markdown", "~> 2.1"
+
+# Add state machine for forms
+gem "aasm", "~> 5.5"
+gem "after_commit_everywhere", "~> 1.0"
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -8,7 +8,7 @@ class Condition < ApplicationRecord
   has_one :form, through: :routing_page
 
   def save_and_update_form
-    save
+    save!
     form.update!(question_section_completed: false)
     form.draft_new_live_form! if form.live?
     form.create_draft_from_archived_form! if form.archived?
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -8,7 +8,10 @@ class Condition < ApplicationRecord
   has_one :form, through: :routing_page
 
   def save_and_update_form
-    save && form.update!(question_section_completed: false)
+    save
+    form.update!(question_section_completed: false)
+    form.draft_new_live_form! if form.live?
+    form.create_draft_from_archived_form! if form.archived?
   end
 
   def destroy_and_update_form!
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -17,14 +17,6 @@ class Form < ApplicationRecord
     pages&.first&.id
   end
 
-  def make_live!(live_at = nil)
-    live_at ||= Time.zone.now
-    touch(time: live_at)
-
-    form_blob = snapshot(live_at:)
-    made_live_forms.create!(json_form_blob: form_blob.to_json, created_at: live_at)
-  end
-
   def has_draft_version
     return true if made_live_forms.blank?
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -44,7 +44,7 @@ class Form < ApplicationRecord
   end
 
   def live_version
-    raise ActiveRecord::RecordNotFound if made_live_forms.blank?
+    raise ActiveRecord::RecordNotFound if made_live_forms.blank? || archived?
 
     made_live_forms.last.json_form_blob
   end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -1,4 +1,7 @@
+require_relative "../state_machines/form_state_machine"
 class Form < ApplicationRecord
+  include FormStateMachine
+
   has_paper_trail
 
   has_many :pages, -> { order(position: :asc) }, dependent: :destroy
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -18,9 +18,7 @@ class Form < ApplicationRecord
   end
 
   def has_draft_version
-    return true if made_live_forms.blank?
-
-    updated_at > live_at
+    draft? | draft_live? || draft_archived?
   end
 
   def draft_version
@@ -32,7 +30,7 @@ class Form < ApplicationRecord
   end
 
   def has_live_version
-    made_live_forms.present?
+    !draft?
   end
 
   def live_version
--- a/app/state_machines/form_state_machine.rb
+++ b/app/state_machines/form_state_machine.rb
@@ -32,16 +32,20 @@ module FormStateMachine
       end
 
       event :make_draft_changes_live do
-        transitions from :draft_live, to: :live
+        transitions from [:draft_live, :draft_archived], to: :live
       end
 
       event :archive_live_form do
         # TODO: Is this only if a form is live and no draft or could a live form be archived even if there is an existing draft?
-        transitions from :live, to: :archived
+        transitions from [:live, :draft_live], to: :archived, guard: Proc.new { has_live_version }
       end
 
       event :create_draft_from_archived_form do
         transitions from: :archived, to: :draft_archived
+      end
+
+      event :make_archived_form_live do
+        transitions from: :archived, to: :live
       end
 
     end
--- a/app/state_machines/form_state_machine.rb
+++ b/app/state_machines/form_state_machine.rb
@@ -9,7 +9,7 @@ module FormStateMachine
       live: "live",
       draft_live: "draft_live",
       archived: "archived",
-      draft_archived: "draft_archived"
+      draft_archived: "draft_archived",
     }
 
     aasm column: :state, enum: true do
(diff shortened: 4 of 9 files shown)

### CODE REVIEW

File: app/state_machines/form_state_machine.rb [hunk 1]
```diff
@@ -0,0 +1,29 @@
+module FormStateMachine
+  extend ActiveSupport::Concern
+
+  included do
+    include AASM
+
+    enum :state, {
+      draft: "draft",
+      live: "live",
```
aliuk2012, line 9:
```suggestion
      initial_draft: "initial_draft",
      live: "live",
      draft: "draft"
```

I was wondering if would be be better than just draft/live because we can track how many forms are yet to be made live at all.

```
Form.initial_draft # returns all Forms that have never been made live

Form.live # returns all Forms that are made live

Form.draft # returns Forms that have changes since the form went live and are waiting for the changes to be made live
```

app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,29 @@
...
+    aasm column: :state, enum: true do
+      state :draft, initial: true
+      state :live
+
+      event :make_form_live do
+        transitions from: :draft, to: :live, guard: Proc.new { task_status_service.mandatory_tasks_completed? }
+      end
+
+      event :make_changes_live do
+        transitions from: :live, to: :draft, guard: Proc.new { has_live_version }
```
aliuk2012, line 21: we do have business rules split across forms-admin and forms-api. For example in forms-admin, if live forms submission email changes when changes are made live then we notify the original submission email address.

Gemfile
```diff
@@ -83,3 +83,7 @@ group :test do
 end
 
 gem "reverse_markdown", "~> 2.1"
+
+# Add state machine for forms
+gem "aasm", "~> 5.5"
+gem "after_commit_everywhere", "~> 1.0"
```
chao-xian, line 89: Why is this needed?

aliuk2012: The `after_commit_everywhere` gem is needed as mentioned in aasm readme https://github.com/aasm/aasm#activerecord. It is used to enabling persisting the the objects state when you call the event name as method with a bang. If the bang is not used then the object state is changed but is not saved in the DB

app/state_machines/form_state_machine.rb
danworth, line 21: I'd have expected `:make_changes_live` to transition `from: :draft, to: :live` but its currently the other way around?

aliuk2012: I was trying to reuse the same language we have in the ui but yes you are correct in this instance it should actually be "draft" to "live" and there should be another event that gets triggered when anything on the form/pages/routing conditions changes.

I think having clear language between the states is what we need.

Another way is to think of "Live" forms as just "versions" of a Form. the latest version of the finalised form are published on govuk. the versioning is just a means for us to "make a commit"  of a form in its current state. This way our "Form" model works and "MadeLiveVersions" are snapshots of the form.

app/state_machines/form_state_machine.rb
chao-xian: * new
* live
* draft
?

app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,29 @@
...
+      draft: "draft",
+      live: "live",
+    }
+
+    aasm column: :state, enum: true do
+      state :draft, initial: true
+      state :live
+
+      event :make_form_live do
+        transitions from: :draft, to: :live, guard: Proc.new { task_status_service.mandatory_tasks_completed? }
```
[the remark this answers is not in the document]
chao-xian: Nice!

app/state_machines/form_state_machine.rb
aliuk2012: resolving this conversation because the event/state names have changed a lot since this conversation took place.

app/models/condition.rb
```diff
@@ -8,7 +8,10 @@ class Condition < ApplicationRecord
   has_one :form, through: :routing_page
 
   def save_and_update_form
-    save && form.update!(question_section_completed: false)
+    save
+    form.update!(question_section_completed: false)
+    form.draft_new_live_form! if form.live?
+    form.create_draft_from_archived_form! if form.archived?
```
aliuk2012, line 14: not going to need these changes if user triggers the state change themselves and specifically "creates a new draft"

app/models/form.rb
```diff
@@ -41,7 +44,7 @@ def has_live_version
   end
 
   def live_version
-    raise ActiveRecord::RecordNotFound if made_live_forms.blank?
+    raise ActiveRecord::RecordNotFound if made_live_forms.blank? || archived?
```
aliuk2012, line 47: may have to put in some other condition to check if runner or admin is trying to view last live version

app/state_machines/form_state_machine.rb
aliuk2012: resolving this as the state names have changed

app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,60 @@
...
+      draft: "draft",
+      live: "live",
+      draft_live: "draft_live",
+      archived: "archived",
+      draft_archived: "draft_archived",
+    }
+
+    aasm column: :state, enum: true do
+      state :draft, initial: true
+      state :live, :draft_live, :archived, :draft_archived
```
DavidBiddle, line 17: Am I right in thinking these are roughly what the different states actually are? 
draft: Has no live snapshot
live: Matches most recent live snapshot
draft_live: Doesn't match most recent live snapshot
archived: Archived and matches live snapshot
draft_archived: Archived and doesn't match live snapshot

aliuk2012: these are the values that would be stored in the `state` column of the forms table. A form would only ever have 1 state at a time. its not tied to any of the snapshots but you could still use `form.live_version` to get the snapshot (if it exists) in any states where you mentioned `snapshot` above

DavidBiddle: Yeah that makes sense - was just checking whether I roughly understand what those states are meant to mean to us

aliuk2012:
if you testing this locally in a rails console

you can check the state/events for a forms using this little lne

```ruby
f = Form.first #returns form record
{ states: f.aasm.states(permitted:true).map(&:name), events: f.aasm.events(permitted:true).map(&:name), state: f.aasm.current_state} #{:states=>[:archived], :events=>[:archive_live_form], :state=>:live}

```

app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,60 @@
...
+  included do
+    include AASM
+
+    enum :state, {
+      draft: "draft",
+      live: "live",
+      draft_live: "draft_live",
+      archived: "archived",
+      draft_archived: "draft_archived",
+    }
```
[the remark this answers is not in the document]
aliuk2012: we can but we were trying to keep it as close to what users see when they view list of statuses

what do others thing about @DavidBiddle  suggestion? @alphagov/forms-devs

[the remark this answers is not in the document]
lfdebrux: I also had comments about the transition names, but I put them on the other PR 😅 

https://github.com/alphagov/forms-api/pull/422#discussion_r1466117774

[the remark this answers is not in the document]
christophercameron-ixd: Not a developer, but this seems like a good idea. It reads more explicit to me. 
And in terms of what the user sees this is basically how they will see it. They don’t know that it’s a single state - they’ll see it as 2 separate things, a draft AND a live/archived form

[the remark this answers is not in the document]
aliuk2012: @thomasiles @jamie-o-wilkinson  and myself had a chat about this is and  were concerned about about how form-admin might map `live_with_draft` state to two different states for the users but it turns up this shouldn't be a problem because form-api forms model `has_draft_version` and `has_live_version` method which forms-admin uses to output the two different tags. we might want to change this in future but for the time being we happy to go with `live_with_draft` state. see implementation in https://github.com/alphagov/forms-api/pull/422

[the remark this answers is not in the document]
lfdebrux:
```mermaid
stateDiagram-v2

    [*] --> Form

    state Form {
        [*] --> Draft
        Draft --> [*]
        --
        Archived --> Live
        Live --> Archived
    }

    Draft --> Live
```

app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,60 @@
...
+  included do
+    include AASM
+
+    enum :state, {
+      draft: "draft",
+      deleted: "deleted",
+      live: "live",
+      draft_live: "draft_live",
+      archived: "archived",
+      draft_archived: "draft_archived",
```
aliuk2012, line 13:
```suggestion
      archived: "archived",
      archived_with_draft: "archived_with_draft",
```

app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    aasm column: :state, enum: true do
+      state :draft, initial: true
+      state :deleted, :live, :draft_live, :archived, :draft_archived
```
aliuk2012, line 18:
```suggestion
      state :deleted, :live, :draft_live, :archived, :archived_with_draft
```

PR #422, opened 2024-01-25
add state column to form model and bare bones state machine

aliuk2012:
### Related 
- https://github.com/alphagov/forms-api/pull/414

This PR adds a state column (but doesn't populate it) and introduces a basic state machine which is not integrated into the forms model just yet.

See individual commit messages when reviewing

```mermaid
---
 title: Form states and events called to transition to state
---
stateDiagram-v2
    Draft --> Live : make_live!
    Draft --> Deleted : delete_form!
    live_with_draft: Live with draft
    Live --> live_with_draft : create_draft_from_live_form!
    live_with_draft --> Live : make_live!
   
```

Trello card:  https://trello.com/c/YTzyf7wz/1339-implement-better-state-of-forms-tracking

### DIFF
--- a/app/state_machines/form_state_machine.rb
+++ b/app/state_machines/form_state_machine.rb
@@ -1,0 +1,49 @@
+module FormStateMachine
+  extend ActiveSupport::Concern
+
+  included do
+    include AASM
+
+    enum :state, {
+      draft: "draft",
+      deleted: "deleted",
+      live: "live",
+      live_with_draft: "live_with_draft",
+    }
+
+    aasm column: :state, enum: true do
+      state :draft, initial: true
+      state :deleted, :live, :live_with_draft
+
+      event :delete_form do
+        after do
+          destroy!
+        end
+
+        transitions from: :draft, to: :deleted
+      end
+
+      event :make_live do
+        after do
+          live_at ||= Time.zone.now
+          touch(time: live_at)
+
+          form_blob = snapshot(live_at:)
+          made_live_forms.create!(json_form_blob: form_blob.to_json, created_at: live_at)
+        end
+
+        transitions from: %i[draft live_with_draft], to: :live, guard: proc { task_status_service.mandatory_tasks_completed? }
+      end
+
+      event :create_draft_from_live_form do
+        transitions from: :live, to: :live_with_draft
+      end
+    end
+
+  private
+
+    def task_status_service
+      @task_status_service ||= TaskStatusService.new(form: self)
+    end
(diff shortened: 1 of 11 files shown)


app/state_machines/form_state_machine.rb
```diff
@@ -0,0 +1,49 @@
...
+          touch(time: live_at)
+
+          form_blob = snapshot(live_at:)
+          made_live_forms.create!(json_form_blob: form_blob.to_json, created_at: live_at)
+        end
+
+        transitions from: %i[draft draft_live], to: :live, guard: proc { task_status_service.mandatory_tasks_completed? }
+      end
+
+      event :draft_new_live_form do
```
lfdebrux, line 38: Can I suggest we call this `create_draft` to match the language used in the 'Create a draft to edit' button?

aliuk2012: Its open for change and I dont have strong opinions on the event names.

🤔 , could this be confusing to think a new form is  being created because "draft" is the initial state?

aliuk2012: related https://github.com/alphagov/forms-api/pull/414#discussion_r1465099113

DavidBiddle: I'm in favour of this suggestion on the same reasoning as the comment on that other PR - it makes sense to me for the transition name to match how the user sees the transition. Not set on it though Al so happy to leave it up to you!

I do agree that there may be some confusion, but I think that confusion is probably unavoidable because the draft/live logic in our code doesn't match the way we present that logic to users. In the longer term I thinik we probably need to resolve that difference, but I appreciate that's much bigger than this PR

aliuk2012: @lfdebrux @DavidBiddle  what about `create_draft_from_live_form`? We now have `live_with_draft` as a status

aliuk2012: other potentials options
- `make_changes_to_live_form`
- `make_draft_from_live_form`
- `add_draft` (not sure about this one makes it sounds like we are creating another new form record)
-

DavidBiddle: I like create_draft_from_live_form/make_draft_from_live_form

lfdebrux: `create_draft_from_live_form` works for me.

DavidBiddle (approved):
Looks good and no issues locally :tada:

PR #427, opened 2024-01-26
Use rails path helpers in specs instead of hard coding them

aliuk2012:

I spotted a few places where we were not using rails path helpers. I've checked a using find `got ` , `post `, `put ` & `delete `

Trello card: 


lfdebrux:
Do we want to change https://github.com/alphagov/forms-api/blob/main/spec/requests/heartbeat_controller_spec.rb#L6 and https://github.com/alphagov/forms-api/blob/main/spec/requests/heartbeat_controller_spec.rb#L17 as well?

lfdebrux (approved):

PR #425, opened 2024-01-25
Integrate FormStateMachine into Form model and start tracking state through events

aliuk2012:
### Related
- https://github.com/alphagov/forms-api/pull/422

### BLOCKED BY 
- https://github.com/alphagov/forms-api/pull/428

### What problem does this pull request solve?

- Adds State Machine to Form model
- Updates make_live action to use `make_live!` state machine event
- Update update methods for form/pages/conditions to transition form to `live_with_draft` if form is live and any of those records are updated
- migrate existing forms to set their state to specific states. (TODO)

Deployed to dev and tested form states were updated to their "inferred" states and new forms states could be added/updated through each event state transition.

Trello card:  https://trello.com/c/YTzyf7wz/1339-implement-better-state-of-forms-tracking

### Things to consider when reviewing

- Ensure that you consider the wider context.
- Does it work when run on your machine?
- Is it clear what the code is doing?
- Do the commit messages explain why the changes were made?
- Are there all the unit tests needed?
- Has all relevant documentation been updated?

PR #428, opened 2024-01-26
tidy up migrations into a single migration file

aliuk2012:
it is good to periodically clean up migration files because it can be difficult to rebuild the db from scratch when you start implementing new required columns etc. and have data migrations running inbetween. We also had a mix of data migration and schema migrations. I reviewed the data migrations and none of them look so important that we would need to keep them as the existing data has already run these. They were also created in a time when we didn't have a way to run rake tasks in AWS.

We already had one issue with timestamps which was fixed here https://github.com/alphagov/forms-api/commit/28b81d344f668cb922054a216e60bdcb023bdccc while trying to add a state machine to the forms model it was picked up in the spike and the final implementation that data migrations were causing an issue because it was expecting a state column in the data migration which hadn't yet been created but was declared in Rails model

https://github.com/alphagov/forms-api/pull/414
https://github.com/alphagov/forms-api/pull/425

I followed this guide https://naturaily.com/blog/remove-old-migrations-ruby-on-rails

What I did to create this
- copy all create/adds in `schema.rb` 
- create a new migration and paste everything copied from the previous step
- remove `force: cascade` from each `create_table`, if this step is ignored then the migration will delete the table and existing data if it exists.
- make a note of the previous migration timestamp.
- delete all other migrations
- rename new migrations timestamp (this is so it shouldn't ran again if the user was kept up to data with migrations.

I deployed this to dev on Saturday and the build/deploy worked without any issues and existing data was available.

Trello card: 


lfdebrux:
I don't think we need to add [db/migrate/20240124160123_create_structure.rb]; Rails seems to be clever enough to handle us deleting old migrations without needed to create a new one [[1](https://guides.rubyonrails.org/active_record_migrations.html#old-migrations)].

Personally I'd prefer not adding a migration file to do nothing, even if we are very sure it won't break anything. It feels like it's just duplicating the purpose of `db/schema.sql`.

aliuk2012:
@lfdebrux agree, I like not having a "base" migration at all but as we proved our correct CI falls over because the tables are not created using `bin/rails db:schema:load` but we agreed to work on that as part of another ticket which @thomasiles will be creating.

PR #429, opened 2024-01-27
Change has_draft_version & has_live_version to use form state column

aliuk2012:
### BLOCKED BY
- https://github.com/alphagov/forms-api/pull/425 needs to be deployed first

This is the last piece of the puzzle, with these changes forms-admin & forms-runner will be able to identify a form's status by the state column. Eventually, we will want to remove these methods but we might want to keep them around as it does allow forms-api to control which statuses should be considered draft or live. forms-admin uses these methods to show/hide the draft/live status tags to users.

Once this is merged we can look at adding Archive state 

Trello card: https://trello.com/c/YTzyf7wz/1339-implement-better-state-of-forms-tracking

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -17,9 +17,7 @@ class Form < ApplicationRecord
   end
 
   def has_draft_version
-    return true if made_live_forms.blank?
-
-    updated_at > live_at
+    draft? || live_with_draft?
   end
 
   def draft_version
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -29,11 +29,11 @@ class Form < ApplicationRecord
   end
 
   def has_live_version
-    made_live_forms.present?
+    live? || live_with_draft?
   end
 
   def live_version
-    raise ActiveRecord::RecordNotFound if made_live_forms.blank?
+    raise ActiveRecord::RecordNotFound unless has_live_version
 
     made_live_forms.last.json_form_blob
   end
--- a/spec/lib/tasks/forms_admin_spec.rb
+++ b/spec/lib/tasks/forms_admin_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe "forms_admin.rake" do
       Rake::Task["forms_admin:make_unlive"].invoke(form.id)
 
       form.reload
-      expect(form.has_live_version).to be false
+      expect(form.made_live_forms.present?).to be false
     end
 
     it "does not make a form unlive if it has no live version" do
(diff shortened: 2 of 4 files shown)


SamJamCul (approved):

app/models/form.rb
```diff
@@ -17,9 +17,7 @@ def start_page
   end
 
   def has_draft_version
-    return true if made_live_forms.blank?
-
-    updated_at > live_at
+    !live?
```
lfdebrux, line 20: I think it would be better to be more explicit with this logic; while it is true currently that there are only three states and only one of those doesn't have a draft, that won't always be true. I think listing out the states that have draft versions will make it easier to update this logic in future.

```suggestion
    draft? || draft_with_live?
```

app/models/form.rb
```diff
@@ -31,11 +29,11 @@ def live_at
   end
 
   def has_live_version
-    made_live_forms.present?
+    !draft?
```
lfdebrux, line 32: As above.

```suggestion
    live? || live_with_draft?
```

spec/models/form_spec.rb
```diff
@@ -175,20 +175,21 @@
...
+      expect(form.has_draft_version).to eq(true)
     end
 
-    it "returns true if form has been made live and been edited" do
-      live_form.update!(name: "Form (edited)")
-
-      expect(live_form.has_draft_version).to eq(true)
+    it "returns false if form is live and no edits" do
+      form.state = :live
+      expect(form.has_draft_version).to eq(false)
```
lfdebrux, line 192: Ideally we would keep the tests as unchanged as possible in the commits changing the method implementation, so we can be confident that the code changes result in the same behaviour.

Can we split the refactor of the tests into a separate commit?

spec/lib/tasks/forms_admin_spec.rb
```diff
@@ -15,7 +15,7 @@
       Rake::Task["forms_admin:make_unlive"].invoke(form.id)
 
       form.reload
-      expect(form.has_live_version).to be false
+      expect(form.made_live_forms.present?).to be false
```
lfdebrux, line 18: It looks like forms that we make unlive with the Rake task/`make_unlive` endpoint will stay in the `live` state now. Although forms that were made unlive change from `draft` to `live` state before #425 was merged should stay in the `draft` state at least.

spec/models/form_spec.rb
aliuk2012: Done, thanks for the suggestion. Its easier to follow the changes now.

PR #430, opened 2024-01-30
add `archive_live_form` state machine event to transition live form to draft

aliuk2012:

This adds a new event to state machine called `archive_live_form!` which for the time being will transition a live/live_with_draft form back to initial `draft` state. This is because the current `make_unlive` form method deletes all `made_live_forms` records and leaves the form in an invalid state of `live`. I think the event can stay as part of archive but when we come to do the archive state we can change the transitions in this event to point to archive/archive_with_draft state

### Before
```mermaid
---
 title: Form states and events called to transition to state
---
stateDiagram-v2
    Draft --> Live : make_live!
    Draft --> Deleted : delete_form!
    live_with_draft: Live with draft
    Live --> live_with_draft : create_draft_from_live_form!
    live_with_draft --> Live : make_live!
   
```

### After
```mermaid
---
 title: Form states and events called to transition to state
---
stateDiagram-v2
    Draft --> Live : make_live!
    Draft --> Deleted : delete_form!
    live_with_draft: Live with draft
    Live --> live_with_draft : create_draft_from_live_form!
    live_with_draft --> Live : make_live!
    live_with_draft --> Draft: archive_live_form!
    Live --> Draft: archive_live_form!
   
```

Trello card: https://trello.com/c/YTzyf7wz/1339-implement-better-state-of-forms-tracking

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -81,6 +81,7 @@ class Form < ApplicationRecord
   delegate :task_statuses, to: :task_status_service
 
   def make_unlive!
+    archive_live_form!
     made_live_forms.destroy_all
   end
 
--- a/app/state_machines/form_state_machine.rb
+++ b/app/state_machines/form_state_machine.rb
@@ -38,6 +38,16 @@ module FormStateMachine
       event :create_draft_from_live_form do
         transitions from: :live, to: :live_with_draft
       end
+
+      event :archive_live_form do
+        # TODO: Temporary transition until we do archive state properly so that live forms cant
+        # be accessed but still exist in the database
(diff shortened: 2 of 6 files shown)


chao-xian:
Can these diagrams be included in the ADR we discussed yesterday?

chao-xian (dismissed):

lfdebrux (dismissed):

lfdebrux:
Can we now update the test in https://github.com/alphagov/forms-api/blob/main/spec/lib/tasks/forms_admin_spec.rb#L11-L19 to show that `form.has_live_version is false`?

PR #435, opened 2024-02-01
[WIP] Add payment_link as an answer type

chao-xian:

Trello card: 

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -9,7 +9,7 @@ class Page < ApplicationRecord
   has_many :goto_conditions, class_name: "Condition", foreign_key: "goto_page_id", dependent: :nullify
   acts_as_list scope: :form
 
-  ANSWER_TYPES = %w[number address date email national_insurance_number phone_number selection organisation_name text name].freeze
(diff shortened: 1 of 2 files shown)


app/models/page.rb
```diff
@@ -9,7 +9,7 @@ class Page < ApplicationRecord
   has_many :goto_conditions, class_name: "Condition", foreign_key: "goto_page_id", dependent: :nullify
   acts_as_list scope: :form
 
-  ANSWER_TYPES = %w[number address date email national_insurance_number phone_number selection organisation_name text name].freeze
+  ANSWER_TYPES = %w[number address date email national_insurance_number phone_number selection organisation_name text name payment_link].freeze
```
aliuk2012, line 12: why is this an answer type? I thought it would have just been another attribute on the form model like "what happens next", declaration and support details.

chao-xian: I've literally no idea. Have been chatting to Sam about it (not freezing you out, but we need to get used to you not being around to answer questions).

aliuk2012: 💯 no probs, just thought i would ask the question. 😄 👍

PR #436, opened 2024-02-01
Remove page_order column from forms table

aliuk2012:
This column is obsolete and was created before Jan 2023 when forms-api was not using rails. It was initial created to start an array of page ids in a particular order but when we moved over to rails we switched over to using acts_as_a_list gem which handled positing and ordering of associated records. I checked in forms-admin adn forms- runner and none of the code uses it nor does any newish form include it.
Trello card: 


thomasiles (approved):
nice one

PR #439, opened 2024-02-05
1342 add archived states to forms state machine

SamJamCul:

https://trello.com/c/QXtBIrKc/1342-add-archived-states-to-forms-state-machine

Adds an `archived` and an `archived_with_draft` state. 
Adds an `archive` endpoint to the forms controller, which will attempt to transition a form into an archived state.
Removes the `make_unlive` endpoint, and maybe the rake task?

Should we refactor the make_unlive task into one that archives a form, or do we no longer want it?


app/state_machines/form_state_machine.rb
```diff
@@ -40,13 +42,12 @@ module FormStateMachine
       end
 
       event :archive_live_form do
-        # TODO: Temporary transition until we do archive state properly so that live forms cant
-        # be accessed but still exist in the database
-        # https://trello.com/c/QXtBIrKc/1342-add-archived-states-to-forms-state-machine
-        transitions from: %i[live live_with_draft], to: :draft
+        after do
+          made_live_forms.destroy_all
+        end
```
aliuk2012, line 47: we dont want to delete live forms, the form state will control `live_version` method so it either serves the live version or it doesn't if the forms state is archived

app/controllers/api/v1/forms_controller.rb
```diff
@@ -61,6 +61,13 @@ def make_unlive
     end
   end
 
+  def archive
+    form.archive_live_form!
+    render json: form, status: :ok
+  rescue AASM::InvalidTransition
+    render json: { error: "Form cannot be archived" }, status: :forbidden
+  end
```
aliuk2012, line 60: I was expecting this to replace the existing make_unlive, make_unlive is not being used in forms-admin or forms-runner

app/state_machines/form_state_machine.rb
SamJamCul: Cool, changed it to achieve this

spec/requests/api/v1/forms_controller_spec.rb
```diff
@@ -404,4 +392,66 @@
...
+      expect(response.status).to eq(200)
+      expect(response.headers["Content-Type"]).to eq("application/json")
+
+      expect(json_body.to_json).to eq made_live_form.json_form_blob
+    end
+
+    it "returns 404 if form has never existed" do
+      get archived_form_path(1), as: :json
+
+      expect(response.status).to eq(404)
```
aliuk2012, line 453:
```suggestion
      expect(response).to have_http_status(:not_found)
```

spec/requests/api/v1/forms_controller_spec.rb
```diff
@@ -404,4 +392,66 @@
...
+      expect(response.headers["Content-Type"]).to eq("application/json")
+      expect(json_body).to eq(error: "Form cannot be archived")
+    end
+
+    context "when the form is live" do
+      it "archives the form" do
+        form = create(:made_live_form).form
+        post archive_form_path(form), as: :json
+
+        expect(response.status).to eq(200)
```
aliuk2012, line 418:
```suggestion
        expect(response).to have_http_status(:ok)
```

spec/requests/api/v1/forms_controller_spec.rb
```diff
@@ -404,4 +392,66 @@
...
+      end
+    end
+
+    context "when the form is live with draft" do
+      it "archives the form with draft" do
+        form = create(:made_live_form).form
+        form.create_draft_from_live_form!
+        post archive_form_path(form), as: :json
+
+        expect(response.status).to eq(200)
```
aliuk2012, line 430:
```suggestion
        expect(response).to have_http_status(:ok)
```

app/controllers/api/v1/forms_controller.rb
```diff
@@ -69,6 +67,10 @@ def show_draft
     render json: form.draft_version, status: :ok
   end
 
+  def show_archived
+    render json: form.last_live_version, status: :ok
```
aliuk2012, line 71: What happens if the form is not archived?

SamJamCul: Not sure what should happen, any suggestions? Could just 404, there's no archived form by that ID etc.

aliuk2012: 404 seems sensible, there is no archived resource for the given form id

spec/requests/api/v1/forms_controller_spec.rb
SamJamCul: sorted

app/controllers/api/v1/forms_controller.rb
SamJamCul: sorted, added a 404 case when the form isn't archived

app/controllers/api/v1/forms_controller.rb
```diff
@@ -52,13 +52,11 @@ def make_live
...
-      form.make_unlive!
-      render json: form, status: :ok
-    else
-      render json: { error: "Form has no live version" }, status: :bad_request
-    end
+  def archive
+    form.archive_live_form!
+    render json: form, status: :ok
+  rescue AASM::InvalidTransition
+    render json: { error: "Form cannot be archived" }, status: :forbidden
```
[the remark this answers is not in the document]
SamJamCul: this should do https://github.com/alphagov/forms-api/pull/439/files#diff-d41344b2d91fe0d353af22f436954a81af57fd0f48cae9642cb8bb2142677c99R403-R410

spec/state_machines/form_state_machine_spec.rb
```diff
@@ -69,17 +69,19 @@ class FakeForm < Form
...
+      let(:form) { FakeForm.new(state: :live) }
 
-    it "transitions to draft if form is live" do
-      expect(form).to transition_from(:live).to(:draft).on_event(:archive_live_form)
+      it "transitions to archived" do
+        expect(form).to transition_from(:live).to(:archived).on_event(:archive_live_form)
+      end
     end
 
     context "when form is live_with_draft" do
```
aliuk2012, line 80: I think we should add an unhappy path here of a draft form calling this event. It should raise an error saying it cant transition to state

SamJamCul: I've added it, I'm a bit wary of doing too many unhappy state change paths as there can be so many!

SamJamCul: https://github.com/alphagov/forms-api/pull/439/files#diff-e80deed9471f0979ccc4d5b60fe3e503adcafd7466d9e312ebffa7ad8bf093fbR72-R78

app/models/form.rb
```diff
@@ -32,12 +32,22 @@ def has_live_version
     live? || live_with_draft?
   end
 
+  def has_been_archived?
```
aliuk2012, line 35: does this method get included in the json to forms-admin and does it get added to the snapshot. Just wondering why this ends with a question mark but `has_draft_version` and `has_live_version` doesn't

aliuk2012 (approved):

PR #443, opened 2024-02-08
Add payment_url to forms

SamJamCul:

https://trello.com/c/VEe72eLB/1357-add-backend-code-to-support-saving-payment-links-for-a-form

Adds a column for `payment_url` to forms. This will be used to store a payment link if one is provided from admin - and to trigger and generate the payment page in the runner

### DIFF
--- a/app/service/task_status_service.rb
+++ b/app/service/task_status_service.rb
@@ -20,6 +20,7 @@ class TaskStatusService
       pages_status:,
       declaration_status:,
       what_happens_next_status:,
+      payment_link_status:,
       privacy_policy_status:,
       support_contact_details_status:,
       make_live_status:,
@@ -58,6 +59,12 @@ private
     else
       :not_started
     end
+  end
+
+  def payment_link_status
+    return :completed if @form.payment_url.present?
+
+    :optional
   end
 
   def privacy_policy_status
--- a/db/migrate/20240208124208_add_payment_url_to_forms.rb
+++ b/db/migrate/20240208124208_add_payment_url_to_forms.rb
@@ -1,0 +1,5 @@
+class AddPaymentUrlToForms < ActiveRecord::Migration[7.1]
+  def change
+    add_column :forms, :payment_url, :string
+  end
+end
--- a/db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
+++ b/db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
@@ -1,0 +1,18 @@
+class AddPaymentUrlToMadeLiveForms < ActiveRecord::Migration[7.1]
+  def change
+    reversible do |direction|
+      MadeLiveForm.find_each do |made_live_form|
+        form_blob = JSON.parse(made_live_form.json_form_blob, symbolize_names: true)
+
+        direction.up do
+          form_blob[:payment_url] = nil
+        end
+        direction.down do
+          form_blob.delete(:payment_url)
+        end
+
+        made_live_form.update!(json_form_blob: form_blob.to_json)
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,9 @@ gem "lograge"
 gem "sentry-rails"
 gem "sentry-ruby"
 
+# Use validate_url so we don't have to write custom URL validation
+gem "validate_url"
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri mingw x64_mingw]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -212,6 +212,7 @@ GEM
     pg (1.5.5)
     psych (5.1.2)
       stringio
+    public_suffix (5.0.4)
     puma (6.4.2)
       nio4r (~> 2.0)
     racc (1.7.3)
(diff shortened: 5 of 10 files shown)


aliuk2012:
did you want to maybe add this column to existing made_live_versions json blob?

DavidBiddle (dismissed):
Looks good! 🎉

db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
```diff
@@ -0,0 +1,18 @@
+class AddPaymentUrlToMadeLiveForms < ActiveRecord::Migration[7.1]
+  def change
+    reversible do |direction|
+      MadeLiveForm.find_each do |made_live_form|
```
chao-xian, line 4: Could you stub an inline model here please: https://jtway.co/data-migrations-with-rails-6fa3d66f128f

db/migrate/20240208124208_add_payment_url_to_forms.rb
```diff
@@ -0,0 +1,5 @@
+class AddPaymentUrlToForms < ActiveRecord::Migration[7.1]
+  def change
+    add_column :forms, :payment_url, :string
```
chao-xian, line 3: Should add a validation that it's a URL?

chao-xian (changes requested):
Couple of comments - otherwise looking good so far!

db/migrate/20240208124208_add_payment_url_to_forms.rb
SamJamCul: I've added this in the model, brought in the `validate_url` gem which we're also using in admin

db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
SamJamCul: Done, I think this is the structure used

chao-xian (dismissed):
Thanks @SamJamCul

DavidBiddle (approved):

DavidBiddle (dismissed):

lfdebrux (approved):
Works for me locally, one minor comment but I think we can merge this without changes 👍

db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
```diff
@@ -0,0 +1,22 @@
+class AddPaymentUrlToMadeLiveForms < ActiveRecord::Migration[7.1]
+  class MadeLiveForm < ApplicationRecord
+    self.table_name = "made_live_forms"
+  end
```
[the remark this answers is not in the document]
SamJamCul: I'm happy to remove it, would it be more appropriate to name the inline model stub as `MadeLiveFormStub`, so that it's disconnected from the actual model properly?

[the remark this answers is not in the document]
lfdebrux: > I'm happy to remove it, would it be more appropriate to name the inline model stub as MadeLiveFormStub, so that it's disconnected from the actual model properly?

I'm not sure 🤔  I mean in this case you're shadowing the existing model, so there's probably no concern about the actual model bleeding through?

Searching for examples in alphagov shows that most other projects don't bother to add the `Stub` suffix, so maybe that's a good guide?

https://github.com/search?q=org%3Aalphagov+path%3A%2F%5Edb%5C%2Fmigrate%5C%2F%2F+ApplicationRecord&type=code

@chao-xian might also have thoughts.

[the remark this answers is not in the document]
chao-xian: Happy for it to be removed too. I did note it but wasn't too fussed at the time. Thanks for mentioning Laurence!

app/service/task_status_service.rb
```diff
@@ -60,6 +61,14 @@ def what_happens_next_status
     end
   end
 
+  def payment_link_status
+    if @form.payment_url.present?
+      :completed
+    else
+      :optional
+    end
```
chao-xian, line 69:
```ruby
return :completed if @form.payment_url.present?

:optional
```

app/service/task_status_service.rb
```diff
@@ -60,6 +61,14 @@ def what_happens_next_status
...
+  end
+
```
chao-xian, line 71:
```suggestion
  def payment_link_status
    return :completed if @form.payment_url.present?
    
    :optional
  end

```

chao-xian: I'm surprised Rubocop didn't suggest this...

chao-xian: Bleurgh... because govuk disabled it https://github.com/alphagov/rubocop-govuk/blob/main/config/style.yml#L125

DavidBiddle: Looks like it's the same throughout this file (not that that's a reason not to change it now)

SamJamCul: sorted!

db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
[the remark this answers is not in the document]
SamJamCul: It's gone!

DavidBiddle (approved):

db/migrate/20240208134335_add_payment_url_to_made_live_forms.rb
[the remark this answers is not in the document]
chao-xian: Sorry I didn't mean for the stub entirely being removed, just the `self.table_name`.

This isn't about solving models bleeding, but because migrations are historical. We'll never remove or rename a class named Forms, but MadeLiveForms is a possibility. 

Having the `class MadeLiveForms < ActiveRecord end;` in a migration stops a migration error.

PR #460, opened 2024-03-20
Set make_live_status for archived form

stephencdaly:

Trello card: https://trello.com/c/UyRNXHGh/1330-implement-archive-a-form-journey

If a form has state `archived`, `make_live_status` was returning a value of nil. This caused an exception to be thrown in forms-admin when loading the page to create a draft version of the form.

Default this method to return "completed" if the form doesn't have a draft version - as this implies that the form must be in either `live` or `archived` state.

### DIFF
--- a/app/service/task_status_service.rb
+++ b/app/service/task_status_service.rb
@@ -88,6 +88,6 @@ private
       return mandatory_tasks_completed? ? :not_started : :cannot_start
     end
 
-    :completed if @form.has_live_version
+    :completed
   end
 end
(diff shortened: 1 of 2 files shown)


stephencdaly:
> Why was this closed? I'm sure there's a good reason but for others to see it as a comment is helpful documentation

I was a bit pre-emptive in raising this one. It's actually linked to a different trello ticket and we need to double check what the behaviour of the task list will be for an archived form. Will re-raise this, or a slightly modified version as appropriate when the next ticket is picked up.

PR #464, opened 2024-03-27
Allow archived forms to be made live

DavidBiddle:

Trello card: https://trello.com/c/li50etsD/1432-implement-form-unarchiving

Updates the state machine to allow archived and archived_with_draft forms to be made live. This will enable us to implement unarchiving a form in forms-admin.

### DIFF
--- a/spec/state_machines/form_state_machine_spec.rb
+++ b/spec/state_machines/form_state_machine_spec.rb
@@ -50,6 +50,30 @@ RSpec.describe FormStateMachine do
         it_behaves_like "transition to live state", FakeForm, :live_with_draft
       end
     end
+
+    context "when form is archived" do
+      let(:form) { FakeForm.new(state: :archived) }
+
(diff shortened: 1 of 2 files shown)


spec/state_machines/form_state_machine_spec.rb
```diff
@@ -39,15 +39,27 @@ class FakeForm < Form
       end
     end
 
-    context "when form is live_with_draft" do
```
stephencdaly, line 42: Have we lost the test for the `live_with_draft` -> `live` transition?

stephencdaly (commented):
Looks good - but has a test mistakenly been removed?

spec/state_machines/form_state_machine_spec.rb
DavidBiddle: Oops, good spot! Have readded

stephencdaly (approved):

PR #465, opened 2024-03-28
Allow user to create a draft for an archived form

stephencdaly:

Trello card: https://trello.com/c/UVCBeTs8/1433-allow-user-to-create-a-draft-for-an-archived-form

This PR:
- Adds a state transition from `archived` -> `archived_with_draft`
- Updates the state of an archived form to `archived_with_draft` when the form or a page of the form is updated
- Returns the appropriate `make_live_status` of the form for an archived form. This will be "Cannot start yet" if there is a draft with incomplete tasks, otherwise it will be "Not started"

You can test this by archiving a form and then clicking "Edit the draft of this form". The task list should show "Not started" for the "Make your changes live" task. If you mark a task as incomplete, this should change to "Cannot start yet".

### DIFF
--- a/app/service/task_status_service.rb
+++ b/app/service/task_status_service.rb
@@ -85,9 +85,11 @@ private
 
   def make_live_status
     if @form.has_draft_version
-      return mandatory_tasks_completed? ? :not_started : :cannot_start
+      mandatory_tasks_completed? ? :not_started : :cannot_start
+    elsif @form.has_been_archived
+      :not_started
+    elsif @form.has_live_version
+      :completed
     end
-
-    :completed if @form.has_live_version
   end
 end
--- a/app/service/task_status_service.rb
+++ b/app/service/task_status_service.rb
@@ -34,31 +34,23 @@ private
   end
 
   def pages_status
-    if @form.question_section_completed && @form.pages.any?
-      :completed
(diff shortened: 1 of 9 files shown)


app/service/task_status_service.rb
```diff
@@ -85,9 +85,11 @@ def support_contact_details_status
 
   def make_live_status
     if @form.has_draft_version
-      return mandatory_tasks_completed? ? :not_started : :cannot_start
```
chao-xian, line 88: Can we refactor this to:

```ruby
  def declaration_status
    return :completed if @form.declaration_section_completed
    return :in_progress if @form.declaration_text.present?

    :not_started
  end
```

https://refactoring.guru/replace-nested-conditional-with-guard-clauses

chao-xian (commented):
Really lovely. Thanks Stephen! I have one suggestion and then I'll approve.

app/service/task_status_service.rb
stephencdaly: I wasn't sure whether you meant this for the method I'd edited or the `declaration_status` method, so I just refactored all the methods

chao-xian: 🤩 thank you!

chao-xian (approved):

PR #509, opened 2024-06-19
Spike add multiple answers

SamJamCul:

https://trello.com/c/pCD4fbNi/1583-spike-work-out-how-to-account-for-routing-and-question-sets

### DIFF
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -1,0 +1,2 @@
+class Question < ApplicationRecord
+end
--- a/app/models/question_set.rb
+++ b/app/models/question_set.rb
@@ -1,0 +1,3 @@
+class QuestionSet < ApplicationRecord
+  has_many :steps
+end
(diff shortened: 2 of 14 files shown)


Gemfile
```diff
@@ -64,6 +64,8 @@ gem "validate_url"
 # For pausing pipelines
 gem "aws-sdk-codepipeline", "~> 1.74"
 
+gem "active_model_serializers"
```
chao-xian, line 67: This is fine for proof of concept but we shouldn't use this gem as it's not fully released (and has gone into maintenance mode) https://github.com/rails-api/active_model_serializers?tab=readme-ov-file#alternatives 

I think we should be able to do something like:
```ruby
JSON.generate(Form.includes(:question_steps).where("form.id = ?", 1).to_h)
```

* https://ruby-doc.org/stdlib-2.0.0/libdoc/json/rdoc/JSON.html#method-i-generate
* https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/QueryMethods.html#includes

stephencdaly: Yeh good point about the gem. I was just scouring the internet for something that we could use to spit out the JSON in the desired format, I'm sure we can find a way without using this particular gem.

PR #538, opened 2024-07-25
Add features report

DavidBiddle:

Trello card: https://trello.com/c/mrozr2YJ/1652-report-for-the-features-okr

- Adds a new ReportController so that the forms-admin can request reports from the API
- Adds a new features report that provides:
  - the total number of live forms
  - the number of live forms that use each answer type at least once
  - the number of pages that use each answer type (across all live forms)
  - the number of live forms with a payment url
  - the number of live forms with at least one route
 
The generated JSON looks like this:
```json
{
  "total_live_forms": 3,
  "live_forms_with_answer_type": {
    "number": 1,
    "address": 1,
    "date": 1,
    "email": 1,
    "national_insurance_number": 1,
    "phone_number": 1,
    "selection": 3,
    "organisation_name": 1,
    "text": 3,
    "name": 1
  },
  "live_pages_with_answer_type": {
    "number": 1,
    "address": 1,
    "date": 1,
    "email": 1,
    "national_insurance_number": 1,
    "phone_number": 1,
    "selection": 4,
    "organisation_name": 2,
    "text": 5,
    "name": 1
  },
  "live_forms_with_payment": 1,
  "live_forms_with_routing": 2
}
```

You can test it locally at [http://localhost:9292/api/v1/reports/features](http://localhost:9292/api/v1/reports/features) (or whatever localhost port your API uses)

We'll need a separate PR in forms-admin to consume this data and display it to the user.

### DIFF
--- a/app/service/features_report_service.rb
+++ b/app/service/features_report_service.rb
@@ -1,37 +1,37 @@
 class FeaturesReportService
   def report
     {
-      total_forms: forms.count,
-      answer_types: { name: number_of_forms_with_answer_type("name"),
-                      organisation_name: number_of_forms_with_answer_type("organisation_name"),
-                      phone_number: number_of_forms_with_answer_type("phone_number"),
-                      email: number_of_forms_with_answer_type("email"),
-                      address: number_of_forms_with_answer_type("address"),
-                      national_insurance_number: number_of_forms_with_answer_type("national_insurance_number"),
-                      date: number_of_forms_with_answer_type("date"),
-                      number: number_of_forms_with_answer_type("number"),
-                      selection: number_of_forms_with_answer_type("selection"),
-                      text: number_of_forms_with_answer_type("text") },
-      payment: number_of_forms_with_payments,
-      routing: number_of_forms_with_routes,
+      total_live_forms: live_forms.count,
+      live_forms_with_answer_type: { name: number_of_live_forms_with_answer_type("name"),
+                                     organisation_name: number_of_live_forms_with_answer_type("organisation_name"),
+                                     phone_number: number_of_live_forms_with_answer_type("phone_number"),
+                                     email: number_of_live_forms_with_answer_type("email"),
+                                     address: number_of_live_forms_with_answer_type("address"),
+                                     national_insurance_number: number_of_live_forms_with_answer_type("national_insurance_number"),
+                                     date: number_of_live_forms_with_answer_type("date"),
+                                     number: number_of_live_forms_with_answer_type("number"),
+                                     selection: number_of_live_forms_with_answer_type("selection"),
+                                     text: number_of_live_forms_with_answer_type("text") },
+      live_forms_with_payment:,
+      live_forms_with_routing:,
     }
   end
 
 private
 
-  def forms
-    @forms ||= Form.all
+  def live_forms
+    @live_forms ||= Form.all.filter(&:has_live_version)
   end
 
-  def number_of_forms_with_answer_type(answer_type)
-    forms.filter { |form| form.pages.any? { |page| page.answer_type == answer_type } }.count
+  def number_of_live_forms_with_answer_type(answer_type)
+    live_forms.filter { |form| form.pages.any? { |page| page.answer_type == answer_type } }.count
   end
 
-  def number_of_forms_with_payments
-    forms.filter { |form| form.payment_url.present? }.count
+  def live_forms_with_payment
+    live_forms.filter { |form| form.payment_url.present? }.count
   end
 
-  def number_of_forms_with_routes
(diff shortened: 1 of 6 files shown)


app/service/features_report_service.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+private
+
+  def live_forms
+    @live_forms ||= Form.all.filter(&:has_live_version)
+  end
+
+  def pages_on_live_forms
+    @pages_on_live_forms ||= live_forms.flat_map(&:pages)
```
stephencdaly, line 19: I wonder how performant this is with loading in all the results and then filtering through them, and whether we'll run into issues holding a large number of forms/pages in memory. I think it's probably better to let the database do a lot of the heavy lifting.

The most performant way to do it would probably be to have some fairly complex SQL queries to extract a lot of the results in one go. But if we don't want to do that, maybe it's better to run a query per result using ActiveRecord query syntax. And then in future if that becomes too slow, we can consider rewriting as raw SQL?

app/service/features_report_service.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def number_of_live_form_pages_with_answer_type(answer_type)
+    pages_on_live_forms.filter { |page| page.answer_type == answer_type }.count
+  end
+
+  def live_forms_with_answer_type
+    Page::ANSWER_TYPES.to_h { |answer_type| [answer_type.to_sym, number_of_live_forms_with_answer_type(answer_type)] }
+  end
+
+  def live_pages_with_answer_type
```
stephencdaly, line 42: I think something like this will work for this one in particular
```
Page.joins(:form).where(forms: {state: 'live'}).group(:answer_type).count
```

app/service/features_report_service.rb
```diff
@@ -0,0 +1,45 @@
...
+  def live_forms_with_routing
+    live_forms.filter { |form| form.pages.any? { |page| page.routing_conditions.present? } }.count
+  end
+
+  def number_of_live_form_pages_with_answer_type(answer_type)
+    pages_on_live_forms.filter { |page| page.answer_type == answer_type }.count
+  end
+
+  def live_forms_with_answer_type
+    Page::ANSWER_TYPES.to_h { |answer_type| [answer_type.to_sym, number_of_live_forms_with_answer_type(answer_type)] }
```
stephencdaly, line 39: Something like this could work for this one, but maybe there's a simpler version
```
Page.joins(:form).where(forms: {state: 'live'}, pages: {answer_type: 'text'}).select('forms.id,pages.answer_type').distinct
```

app/service/features_report_service.rb
DavidBiddle: Wasn't sure how well this would perform on a database with more forms, so good to get your view on this. Thanks for the suggestions, think I've got working ActiveRecord queries for all of them now

app/service/features_report_service.rb
DavidBiddle: I was getting some test failures that I couldn't work out (have figured it out now - it was generating hashes with string keys instead of symbols). It also misses out any answer types that don't have keys but I think that's OK actually, it's more concise and should be able to deal with it in admin without much trouble. Have updated (also managed to find a similar approach for the live_forms_with_answer_type query.

spec/service/features_report_service_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+          response = features_report_service.report
+
+          expect(response[:live_forms_with_routing]).to eq 1
+        end
+      end
+    end
+  end
+
+  context "when the form is not live" do
+    let(:state) { %i[draft archived archived_with_draft].sample }
```
stephencdaly, line 112: Same here, it would take a while for tests to run with all the states, should we create a form of each type instead?

spec/service/features_report_service_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+          (build :page, answer_type: "national_insurance_number"),
+          (build :page, answer_type: "date"),
+          (build :page, answer_type: "number"),
+          (build :page, answer_type: "selection", routing_conditions: [(build :condition)]),
+          (build :page, answer_type: "text"),
+        ]
+      end
+
+      it "does not count the form in any of the report metrics" do
+        response = features_report_service.report
```
stephencdaly, line 133: We are creating the form before this test is run, but it would be quite easy to play with the `before` and make a mistake that would mean this test was passing because no forms exist, rather than them not being live. 

To be safe, could we maybe add in an assertion that the form exists in the database so we're sure the test is passing for the right reason?

stephencdaly (approved):
Excellent work!

PR #542, opened 2024-08-01
Add attribute for "add another answer" for a single question

lfdebrux:

Trello card: https://github.com/alphagov/forms-admin/pull/1364 

We want to be able to support an "add another answer" feature for questions in forms.

This PR allows a single question to be marked as allowing more than one answer, in the simplest possible way.

Currently there is no support for this attribute in forms-admin or forms-runner, so the best way to test this is with an API client such as httpie or postman.

Also note that https://github.com/alphagov/forms-admin/pull/1364 needs to be merged before this PR, otherwise the end to end tests will break as forms-admin will refuse the new attribute.

### DIFF
--- a/app/controllers/api/v1/pages_controller.rb
+++ b/app/controllers/api/v1/pages_controller.rb
@@ -90,6 +90,7 @@ private
       :answer_type,
       :next_page,
       :is_optional,
+      :is_repeatable,
       *guidance_params,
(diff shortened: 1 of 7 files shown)


thomasiles (approved):
This looks good to me ⭐ 

It works locally and I couldn't find a way it disrupted the runner or admin.

PR #544, opened 2024-08-02
1681 3 add unique form url code column to api 3

SamJamCul:

https://trello.com/c/NNdV9jj3/1681-3-add-unique-form-url-code-column-to-api-3

The first part of adding external IDs for forms - this adds the external ID column, and starts assigning them for newly created forms with the their normal ID. 

The next step is to migrate all existing forms into the same format, so that all forms from that point onwards have an external id, which is equivalent to their current ID.

There aren't any UI aspects to this yet, so a little hard to test - but you can, once pulled down, check that all your existing forms are migrated with no external ID - but new forms are given one. 

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -55,13 +55,14 @@ class Form < ApplicationRecord
   end
 
   def as_json(options = {})
+    options[:except] ||= [:external_id]
     options[:methods] ||= %i[live_at start_page has_draft_version has_live_version has_routing_errors ready_for_live incomplete_tasks task_statuses]
     super(options)
   end
 
   def snapshot(**kwargs)
     # override methods so it doesn't include things we don't want
-    as_json(except: :state,
+    as_json(except: %i[state external_id],
             include: {
               pages: {
                 include: {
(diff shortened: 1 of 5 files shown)


app/lib/external_id_provider.rb
```diff
@@ -0,0 +1,5 @@
+class ExternalIdProvider
+  def self.generate_id
+    SecureRandom.base58(8)
+  end
```
lfdebrux, line 4: Seeing as the only place we're using this code in this app currently we're immediately downcasing the output, I'd suggest either changing `generate_id` to also call `downcase`.

If we're keen to preserve compatibility with the code in forms-admin we could instead add a new method `generate_case_insensitive_id` and call that instead.

The current code has the behaviour split between the definition and the callsite, which I found confusing on first reading.

SamJamCul: Happy to change this, I think I also prefer it all living in one place, and the model can trust that the ID generator is giving it something appropriate.

lfdebrux (approved):

PR #546, opened 2024-08-06
Add rake task for setting external_id to id for all forms

SamJamCul:

https://trello.com/c/UUocDXYa/1680-5-migration-for-existing-form-external-ids-5

This rake task will attribute every form's external id as its id. This is meant to be a task that is run once, and after this https://github.com/alphagov/forms-api/pull/544 has been deployed. 

From that point forwards, all new forms and existing forms will have an `external_id`, which will lead on to the next steps for moving away from exposing our form ids

### DIFF
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -12,4 +12,13 @@ namespace :forms do
 
     puts "Updated organisation for Form: #{form.id} to be #{args[:organisation_id]}"
   end
+
+  desc "Sets the external_id for all forms to their id"
+  task set_external_ids: :environment do
+    puts "setting external_id for each form to their id"
+
+    Form.find_each { |form| form.update_column(:external_id, form.id) }
+
+    puts "external_id has been set for each form to their id"
+  end
 end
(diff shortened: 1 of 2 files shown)


lib/tasks/forms.rake
```diff
@@ -12,4 +12,13 @@ namespace :forms do
 
     puts "Updated organisation for Form: #{form.id} to be #{args[:organisation_id]}"
   end
+
+  desc "Sets the external_id for all forms to their id"
+  task set_external_ids: :environment do
+    puts "setting external_id for each form to their id"
+
+    Form.find_each { |form| form.update_columns(external_id: form.id) }
```
jamie-o-wilkinson, line 20: I guess `update_column` here would work and might express that you are only updating a single column more clearly? But it amounts to the same thing.

jamie-o-wilkinson (approved):
I've left a minor comment, but otherwise looks good!

lib/tasks/forms.rake
SamJamCul: Ah good catch, didn't know there was a better variation

SamJamCul: Fixed

jamie-o-wilkinson (approved):

PR #551, opened 2024-08-16
Add rake task to remove access token

chao-xian:

We need to remove a developer's access token. At the moment our workflow relies on another dev with a valid token to follow the steps with our access token API. As an API this makes sense for users, but there are situations where no one has an access token and it's just a bit painful.

We don't want to have to do things like run a SQL query or raiding a backup token.

The intention is to use our cli tool which is restricted.

https://trello.com/c/GDNpsobU

### DIFF
--- a/lib/tasks/access_tokens.rake
+++ b/lib/tasks/access_tokens.rake
@@ -1,0 +1,13 @@
+namespace :access_tokens do
+  desc "Remove an access token for a user"
+  task :remove_access_token, %i[owner_email] => :environment do |_, args|
+    usage_message = "usage: rake access_tokens:remove_access_token[<owner_email>]".freeze
+    abort usage_message if args[:owner_email].blank?
+
+    token = AccessToken.find_by(owner: args[:owner_email])
+    abort "Access token with owner #{args[:owner_email]} not found" unless token
+
+    token.destroy!
+    puts "Access token removed for #{args[:owner_email]}"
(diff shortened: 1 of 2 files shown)


lib/tasks/access_tokens.rake
```diff
@@ -0,0 +1,13 @@
+namespace :access_tokens do
+  desc "Remove an access token for a user"
+  task :remove_access_token, %i[owner_email] => :environment do |_, args|
```
lfdebrux, line 3: This should probably be just `owner` instead of `owner_email`, as we don't normally put the email address in the owner field.

```suggestion
  task :remove_access_token, %i[owner] => :environment do |_, args|
```

chao-xian: Thanks @lfdebrux. What format do the values for `owner` usually take? I had assumed as it's an API token it'd be sensible to make it email addresses...

lfdebrux: At the moment we don't have a consistent format, in the production database we've used `kelvin.gan`, `kelvin_gan` and `kelvin-gan`, but the most common is the kebab case (`kelvin-gan`).

chao-xian: Right well I'm going to ask from this day on we use email addresses :)

lfdebrux (approved):

PR #557, opened 2024-08-29
Add error handling on api token removal rake task

chao-xian:
If something goes wrong we don't necessarily see that it does with how this rake task was written before.

Trello card: https://trello.com/c/GDNpsobU/1650-leaving-checklist-for-dan-%F0%9F%98%AD 


lfdebrux:
I'm concerned about why this is necessary; I thought exceptions should be printed to stderr...

chao-xian:
@lfdebrux Good point! I think I'm being overcautious... will close

PR #567, opened 2024-09-09
Prepare for dropping form.organisation_id column

stephencdaly:

Trello card: https://trello.com/c/ZHGNp5Lk/1774-remove-organisationid-from-forms-api

Forms-admin no longer consumes the `form.organisation_id` - see https://github.com/alphagov/forms-admin/pull/1423.

This PR removes all uses of `form.organisation_id` from the forms-api code. It also sets the `ignored_columns` for the Forms ActiveModel to safely allow the column to be dropped. See the [ActiveRecord documentation](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-ignored_columns-3D) for usage of `ignored_columns`.

This follows the steps outlined in this [blog](https://andycroll.com/ruby/safely-remove-a-column-field-from-active-record/) to safely remove a column.

Once this is merged, a follow-up PR will include a migration to drop the column from the forms table.

### DIFF
--- a/app/controllers/api/v1/forms_controller.rb
+++ b/app/controllers/api/v1/forms_controller.rb
@@ -5,11 +5,9 @@ class Api::V1::FormsController < ApplicationController
   end
 
   def index
-    organisation_id = params[:organisation_id]
     creator_id = params[:creator_id]
 
     forms = Form.all
-    forms = forms.filter_by_organisation_id(organisation_id) if organisation_id.present?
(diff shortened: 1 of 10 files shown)


lfdebrux (approved):
Nice one. I particularly like the clear well described commits!

I've tested this locally and it works for me 🙌

PR #569, opened 2024-09-10
Remove organisation_id column from forms table

stephencdaly:

Trello card: https://trello.com/c/ZHGNp5Lk/1774-remove-organisationid-from-forms-api

This column is no longer referenced, and is ignored by the Forms ActiveRecord, so can be safely dropped.

### DIFF
--- a/db/migrate/20240910115155_remove_organisation_id_from_forms.rb
+++ b/db/migrate/20240910115155_remove_organisation_id_from_forms.rb
@@ -1,0 +1,5 @@
+class RemoveOrganisationIdFromForms < ActiveRecord::Migration[7.1]
+  def change
+    remove_column :forms, :organisation_id, :bigint
+  end
+end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -14,8 +14,6 @@ class Form < ApplicationRecord
 
   after_create :set_external_id
 
-  self.ignored_columns = %w[organisation_id]
-
   def start_page
     pages&.first&.id
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_08_01_123903) do
+ActiveRecord::Schema[7.1].define(version: 2024_09_10_115155) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -52,7 +52,6 @@ ActiveRecord::Schema[7.1].define(version: 2024_08_01_123903) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.bigint "creator_id"
-    t.bigint "organisation_id"
     t.text "what_happens_next_markdown"
     t.string "state"
     t.string "payment_url"


lfdebrux:
Can we also remove the [Form.ignored_columns line](https://github.com/alphagov/forms-api/blob/b6fac9f36ccd2ce2671681dcd769809efbd73efc/app/models/form.rb#L17) in this PR?

db/migrate/20240910115155_remove_organisation_id_from_forms.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveOrganisationIdFromForms < ActiveRecord::Migration[7.1]
+  def change
```
lfdebrux, line 2: I think we probably want to mark this migration as non-reversible? What do you think?

stephencdaly:
> Can we also remove the [Form.ignored_columns line](https://github.com/alphagov/forms-api/blob/b6fac9f36ccd2ce2671681dcd769809efbd73efc/app/models/form.rb#L17) in this PR?

@lfdebrux I think probably yes - but to be extra sure I was going to do it in a subsequent PR. But that's definitely being more cautious than is necessary. Should I just add it in this PR?

db/migrate/20240910115155_remove_organisation_id_from_forms.rb
stephencdaly: If you think it's necessary? Technically it can be reversed, but we do lose the data. Do we always need to make sure we mark migrations as non-reversible if they shouldn't be reversed?

lfdebrux:
> > Can we also remove the [Form.ignored_columns line](https://github.com/alphagov/forms-api/blob/b6fac9f36ccd2ce2671681dcd769809efbd73efc/app/models/form.rb#L17) in this PR?
> 
> @lfdebrux I think probably yes - but to be extra sure I was going to do it in a subsequent PR. But that's definitely being more cautious than is necessary. Should I just add it in this PR?

Oh I see. I think it should be fine to keep them both in the same PR, as the migration should always happen (and succeed) before the code changes are deployed.

db/migrate/20240910115155_remove_organisation_id_from_forms.rb
lfdebrux: In this case it's not technically necessary, as you say, but I think it's not very useful to be able to rollback that migration.

The fact that the Rails guide example for throwing an error to prevent reverts does suggest to me that it's best practice, then again [this stack overflow post](https://stackoverflow.com/questions/621257/rails-is-it-bad-to-have-an-irreversible-migration) suggests it's a bad idea to have an irreversible migration in production code...

stephencdaly: As we don't think anyone would want to roll back that migration because it's not useful, reckon it's fine to leave it as it is for simplicity?

stephencdaly:
> > > Can we also remove the [Form.ignored_columns line](https://github.com/alphagov/forms-api/blob/b6fac9f36ccd2ce2671681dcd769809efbd73efc/app/models/form.rb#L17) in this PR?
> > 
> > 
> > @lfdebrux I think probably yes - but to be extra sure I was going to do it in a subsequent PR. But that's definitely being more cautious than is necessary. Should I just add it in this PR?
> 
> Oh I see. I think it should be fine to keep them both in the same PR, as the migration should always happen (and succeed) before the code changes are deployed.

@lfdebrux added a commit to do this

lfdebrux (approved):

PR #572, opened 2024-09-16
Add submission type to the form model

jamie-o-wilkinson:

This PR adds a submission type to the form model. This will be used by forms-runner to determine how to deliver submission data, and by forms-admin to select which submission type to use.

This will supersede the per-form feature flagging used for CSV submissions, and is required in advance of refactoring CSV submissions to be production ready.


lfdebrux (approved):

db/migrate/20240126213348_set_existing_form_state.rb
```diff
@@ -1,23 +0,0 @@
-class SetExistingFormState < ActiveRecord::Migration[7.1]
```
chao-xian, line 1: If we add in 

```ruby
  class Form < ActiveRecord::Base
    self.table_name = :forms
  end
```

then this no longer fails

stephencdaly: The migration does nothing anyway, and a while ago we wanted to get rid of the `has_live_version` method anyway but I wasn't sure what to do about this migration - so I reckon let's delete anyway.

chao-xian (approved):

PR #573, opened 2024-09-16
Add rake task to update submission type

jamie-o-wilkinson:

This PR adds a rake task to update `submission_type` for a given form. This will be a temporary rake task to set the `submission_type` for forms that currently have CSV submission enabled via the per form feature flag.

### DIFF
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -7,4 +7,13 @@ namespace :forms do
 
     puts "external_id has been set for each form to their id"
   end
+
+  desc "Set submission_type to email_with_csv"
+  task :set_submission_type_to_email_with_csv, %i[form_id] => :environment do |_, args|
+    puts "setting submission_type to email_with_csv for form: #{args[:form_id]}"
+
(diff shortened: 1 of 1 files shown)


lib/tasks/forms.rake
```diff
@@ -7,4 +7,13 @@ namespace :forms do
 
     puts "external_id has been set for each form to their id"
   end
+
+  desc "Set submission_type to email_with_csv"
+  task :set_submission_type_to_email_with_csv, %i[form_id] => :environment do |_, args|
```
DavidBiddle, line 12: Optional: you could put a usage message in here (like the one in access_tokens.rake) to make dealing with errors easier

DavidBiddle (approved):
Code looks good and the task works for me locally :tada: 

Have left one suggestion but it's optional

lib/tasks/forms.rake
stephencdaly: Great suggestion, we'll do that

stephencdaly: done

PR #577, opened 2024-09-18
Add api_v2_form_documents table and model

thomasiles:
### Add Form_document

Trello card: https://trello.com/c/l48nxEEO/1789-8-add-a-form-documents-table-to-forms-api

This PR just covers creating and settling the table structure and model for form documents. It doesn't try to update the new model with either version of the API and doesn't add new endpoints. These will come in other PRs.

To represent `form_documents` which will be used to model the v2 API we need a new model and database table.

This model will hold the 'content' of a form as a JSON structure. A form can have more than one form_document and will be referenced by its tag.

Tags will be used to represent the different states of a form. For example most forms will have a form_document with a tag draft. Forms which have been made live will have a form with the tag 'live' and archived forms will have a form document with the tag 'archived'.

Currently there is no validation on the form_document, other than it has to be present. There is also no index on tag, which might be something to add in the future if needed.

FormDocuments belong to Api_v2_forms rather than the standard Form.

This commit doesn't include the association on Api_v2_forms. That should be added as needed.

### DIFF
--- a/app/models/api/v2.rb
+++ b/app/models/api/v2.rb
@@ -1,0 +1,5 @@
+module Api::V2
+  def self.table_name_prefix
+    "api_v2_"
+  end
+end
--- a/db/migrate/20240916122719_create_api_v2_form_documents.rb
+++ b/db/migrate/20240916122719_create_api_v2_form_documents.rb
@@ -1,0 +1,13 @@
+class CreateApiV2FormDocuments < ActiveRecord::Migration[7.1]
+  def change
+    create_table :api_v2_form_documents do |t|
+      t.references :form, foreign_key: true, comment: "The form this document belongs to"
+      t.text :tag, null: false,  comment: "The tag for the form, for example: 'live' or 'draft'"
+      t.jsonb :content, comment: "The JSON which describes the form"
+
+      t.timestamps
+
+      t.index %i[form_id tag], unique: true, name: "index_api_v2_form_documents_on_form_id_and_tag"
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


app/models/api/v2.rb
```diff
@@ -0,0 +1,5 @@
+module Api::V2
+  def self.table_name_prefix
+    "api_v2_"
+  end
+end
```
thomasiles, line 5: The Rails generator added this when I created the model with the api_v2 prefix. [Normally it's the empty string](https://apidock.com/rails/ActiveRecord/Base/table_name_prefix/class), we use api_v2 here to keep our things separate from the existing form models, following on from the way the controllers and other v2 files are in a module.

db/migrate/20240916122719_create_api_v2_form_documents.rb
```diff
@@ -0,0 +1,11 @@
+class CreateApiV2FormDocuments < ActiveRecord::Migration[7.1]
+  def change
+    create_table :api_v2_form_documents do |t|
+      t.references :form, foreign_key: true, comment: "The form this document belongs to"
+      t.text :tag, null: false,  comment: "The tag for the form, for example: 'live' or 'draft'"
```
thomasiles, line 5: I've called this tag, based on the [ADR](https://github.com/alphagov/forms/blob/main/ADR/ADR034-api-v2-for-forms.md).

I think it could also be called state, but tag seems a bit nicer.

There is no validation on it, other than it must be present. Do we want to restrict the possible values to draft, live and archived or is that something the admin should be in charge of?

lfdebrux: Yep, that makes sense to me. I think we agreed as a team that keeping validation minimal in the API is good for now (at least while the only clients are our own apps).

lfdebrux (approved):
Works for me, nice and simple!

PR #578, opened 2024-09-19
Add submission type to live forms

jamie-o-wilkinson:

This PR adds a `submission_type` attribute to the snapshot JSON of every made live form. This is necessary so that forms-runner can safely use this attribute to determine submission type.

This PR also updates the `set_submission_type_to_email_with_csv` rake task to set the `submission_type` to `email_with_csv` for a form's live versions.

### DIFF
--- a/db/migrate/20240918153233_add_submission_type_to_made_live_forms.rb
+++ b/db/migrate/20240918153233_add_submission_type_to_made_live_forms.rb
@@ -1,0 +1,22 @@
+class AddSubmissionTypeToMadeLiveForms < ActiveRecord::Migration[7.1]
+  class MadeLiveForm < ApplicationRecord
+    self.table_name = :made_live_forms
+  end
+
+  def change
+    reversible do |direction|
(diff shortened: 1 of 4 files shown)


db/migrate/20240918153233_add_submission_type_to_made_live_forms.rb
```diff
@@ -0,0 +1,18 @@
+class AddSubmissionTypeToMadeLiveForms < ActiveRecord::Migration[7.1]
+  def change
+    reversible do |direction|
+      MadeLiveForm.find_each do |made_live_form|
```
chao-xian, line 4: Whenever we use model code directly in our migrations we should stub the class ie ```class MadeLiveForm < ActiveRecord::Base
    self.table_name = :forms
  end``` to avoid any errors in re-running migrations say if validations changed or the model is removed altogether.

stephencdaly: Thanks! Have added this

chao-xian (approved):

PR #592, opened 2024-10-07
Add task to share preview of draft

stephencdaly:

Trello card: https://trello.com/c/QuqK5jmG/1867-add-new-task-to-share-a-preview-of-your-draft-form-for-drafts-of-new-live-and-archived-forms

We are adding a new task to the draft form task list to share a preview of the form. This task will have radio buttons to mark the task as completed.

Add a migration to add a new `share_preview_completed` boolean column to the forms table to store whether the new task has been completed.

Default the value to `false`. This means that if a draft version of an existing live form is created, this task will need to be completed before the draft can be made live.

Add a task_status to the form to represent whether this new task has been completed. This will be a mandatory task, but for now don't add a check to the `incomplete_tasks` method as we need to update the pages for the new task in forms-admin first.


lfdebrux (approved):
Looks good to me, I've tested locally and it works!

PR #596, opened 2024-10-09
Update V2 Form models with V1 API endpoints

thomasiles:
### Create, update and delete FormDocuments when using the V1 API

Trello card: https://trello.com/c/l48nxEEO/1789-8-add-a-form-documents-table-to-forms-api

This PR adds a migration and inserts calls into the V1 API to ensure the V2 FormDocument models are updated with calls to the current, V1 API.

It's tempting to add callbacks into the `Form` and `Page` model to update `FormDocuments` when they are committed to the database. That ends up being less code but feels more likely to cause issues somewhere along the line. Adding the calls to the controller is more inline with our goal - updating FormDocuments in API calls.

The code to keep the FormDocuments in sync is aiming to keep things simple at the cost of making more database calls. Rather than use separate methods for the migration and ongoing updates, it uses only one.

 
### DIFF
--- a/spec/service/api/v2/form_document_sync_service_spec.rb
+++ b/spec/service/api/v2/form_document_sync_service_spec.rb
@@ -1,0 +1,133 @@
+require "rails_helper"
+
+RSpec.describe Api::V2::FormDocumentSyncService do
+  let(:service) { described_class.new }
+  let(:form) { create(:form) }
+  let(:converter) { instance_double(Api::V2::Converter) }
+
+  before do
+    allow(Api::V2::Converter).to receive(:new).and_return(converter)
+    allow(converter).to receive(:to_api_v2_form_document).and_return({})
+  end
+
+  describe "#synchronize_form" do
+    context "when form state is live" do
+      before { allow(form).to receive(:state).and_return("live") }
+
+      it "calls sync_live_form" do
+        expect(service).to receive(:sync_live_form).with(form)
+        service.synchronize_form(form)
+      end
+    end
+
+    context "when form state is live_with_draft" do
+      before { allow(form).to receive(:state).and_return("live_with_draft") }
+
+      it "calls sync_live_with_draft_form" do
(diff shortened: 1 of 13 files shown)


spec/service/api/v2/form_document_sync_service_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+  describe "#delete_form_documents" do
+    it "deletes form documents with specified tags" do
+      form_documents = instance_double(ActiveRecord::Relation)
+      allow(Api::V2::FormDocument).to receive(:where).with(form_id: form.id, tag: %i[draft archived]).and_return(form_documents)
+      allow(form_documents).to receive(:delete_all).and_return(true)
+
+      service.delete_form_documents(form, %i[draft archived])
+
+      expect(Api::V2::FormDocument).to have_received(:where).with(form_id: form.id, tag: %i[draft archived])
+      expect(form_documents).to have_received(:delete_all)
```
lfdebrux, line 130: I think we could rewrite this test to use the database instead of mocking the model:

```suggestion
      create(:api_v2_form_document, form_id: form.id, tag: :draft)
      create(:api_v2_form_document, form_id: form.id, tag: :archived)
      create(:api_v2_form_document, form_id: form.id, tag: :live)
      
      expect {
          service.delete_form_documents(form, %i[draft archived])
      }.to change(Api::V2::FormDocuments, :count).from(3).to(0)
```

That way we're testing the behaviour rather than the implementation.

lfdebrux (dismissed):
Nice one! 👏 

This looks good to me, and works on my machine.

I have one small comment about the tests, but otherwise I think this is good to go.

spec/service/api/v2/form_document_sync_service_spec.rb
thomasiles: nice idea, I've updated it

lfdebrux (approved):

PR #601, opened 2024-10-15
Connect to s3 bucket for form submission

alice-carr:

[Trello card](https://trello.com/c/ToXPf8zb/1869-get-code-into-production-behind-a-feature-flag-for-the-s3-bucket-option)

this pr:
- adds a new column s3_bucket_name to the forms table
- adds a rake task to set the submission_type to s3 and the s3_bucket_name to an s3 bucket name
- adds a new submission type of s3
- updates all tests relating to those things


stephencdaly:
> In addition to the S3 bucket name we should also require the AWS account ID so that we can verify the bucket owner as described in https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-owner-condition.html#bucket-owner-condition-use.

Thanks for the suggestion @james-sheppard-gov-uk, have added that!

lib/tasks/forms.rake
```diff
@@ -77,3 +97,23 @@ def summarise_form_documents
...
 end
+
+def set_submission_type(submission_type, form_id)
+  Rails.logger.info("setting submission_type to #{submission_type} for form: #{form_id}")
+
+  form = Form.find(form_id)
+  form.submission_type = submission_type
+  form.save!
+
+  made_live_form = form.made_live_forms.last
```
lfdebrux, line 108: Would `form.made_live_forms.last` return the old form snapshot if the form was archived? Do we want to update the archived form document?

lib/tasks/forms.rake
```diff
@@ -8,27 +8,47 @@ namespace :forms do
...
-      form_blob[:submission_type] = "email_with_csv"
+      form_blob[:submission_type] = "s3"
+      form_blob[:s3_bucket_name] = args[:s3_bucket_name]
+      form_blob[:s3_bucket_aws_account_id] = args[:s3_bucket_aws_account_id]
 
       made_live_form.update!(json_form_blob: form_blob.to_json)
     end
-
-    Rails.logger.info("set submission_type to email_with_csv for form: #{args[:form_id]}")
+    Rails.logger.info("set submission_type to s3 s3_bucket_name to #{args[:s3_bucket_name]} for form: #{args[:form_id]}")
```
lfdebrux, line 51:
```suggestion
    Rails.logger.info("set submission_type to s3 and s3_bucket_name to #{args[:s3_bucket_name]} for form: #{args[:form_id]}")
```

lib/tasks/forms.rake
stephencdaly: Oh very good point! We can use `form.live_version` instead I think

stephencdaly: Ah not quite actually because we need update it and that method returns the JSON - but we can check the form status

lfdebrux (approved):

PR #607, opened 2024-10-17
Add CSV submission to features report

DavidBiddle:

Trello card: None, followup from the additional submission types work

Adds the number of forms which have CSV submissions enabled to the features report API which admin consumes.

Needed for https://github.com/alphagov/forms-admin/pull/1523


stephencdaly (approved):
Looks good to me, have tested locally and it works as expected. Thanks for adding that comment!

PR #614, opened 2024-10-22
Store s3 bucket region on form

stephencdaly:

Trello card: https://trello.com/c/ToXPf8zb/

When using the AWS SDK for Ruby, it is required to provide the bucket region. If the region is different to the region the bucket was created in, we get the following error:

Aws::s3::Errors::PermanentRedirect - The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.

Add a column `s3_bucket_region` to the forms table to store the region for the S3 bucket to send form submission data to.

### DIFF
--- a/db/migrate/20241022154933_add_s3_bucket_region_to_forms.rb
+++ b/db/migrate/20241022154933_add_s3_bucket_region_to_forms.rb
@@ -1,0 +1,5 @@
+class AddS3BucketRegionToForms < ActiveRecord::Migration[7.2]
+  def change
+    add_column :forms, :s3_bucket_region, :string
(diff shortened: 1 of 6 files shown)


chao-xian (approved):
haven't run the rake task or checked against a bucket but the code looks good - I think we might want to consider a SubmissionType class in future...

stephencdaly:
> haven't run the rake task or checked against a bucket but the code looks good - I think we might want to consider a SubmissionType class in future...

Yeh we're adding quite a lot onto the form now. If/when we revisit this, it might be worthwhile.

PR #615, opened 2024-10-23
Add full form report for add another answer

SamJamCul:
This adds a new report which will return all forms that contain a question that has `is_repeatable` set to true. This includes forms in draft states only, as we want to get an idea of how this feature is being used even in non-live forms.

Trello card: https://trello.com/c/a2bMFMxU/1827-5-report-when-forms-are-using-add-another-answer-single-qn

This adds an additional report field which will give a few details about every form with a repeatable question. It will bundle each form as a combination of its form id, form name, and then every repeatable question within that form. Each question will have its id and question text bundled up as well. 

This is in order to create a minimal report in admin, which can display the name of a form and the questions' texts, as well as provide links to them. 

### DIFF
--- a/app/services/features_report_service.rb
+++ b/app/services/features_report_service.rb
@@ -8,6 +8,7 @@ class FeaturesReportService
       live_forms_with_routing:,
       live_forms_with_add_another_answer:,
       live_forms_with_csv_submission_enabled:,
+      all_forms_with_add_another_answer:,
     }
   end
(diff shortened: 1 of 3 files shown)


thomasiles (approved):
Looks good!

I think the SQL statement is returning all fields for the Form, when only 3 are needed. I don't think this is a performance issue at the moment though.

PR #618, opened 2024-10-25
Add state to all add another answer forms report

SamJamCul:
This adds the state field to the add another answer forms report. This will let us identify what state forms are in with the report, which will help filter out draft forms from live ones.

Trello card: https://trello.com/c/a2bMFMxU/1827-report-when-forms-are-using-add-another-answer-single-qn

Adds a state field to the add another answer report, so that we can more easily sift through all the forms!

### DIFF
--- a/app/services/features_report_service.rb
+++ b/app/services/features_report_service.rb
@@ -54,6 +54,7 @@ private
       {
         form_id: form.id,
         name: form.name,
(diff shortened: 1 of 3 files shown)


stephencdaly (approved):
Looks good, works as expected.

PR #622, opened 2024-10-29
Spike: create new form in seed for e2e testing with submission type s3

alice-carr:

Trello card: [](https://trello.com/c/OJHCEcWx/1937-spike-add-an-e2e-test-for-s3-submissions)

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -7,6 +7,8 @@
 #   Character.create(name: "Luke", movie: movies.first)
 
 submission_email = ENV["EMAIL"] || `git config --get user.email`.strip
+s3_bucket_aws_account_id = ENV["AWS_ACCOUNT_ID"]
+s3_bucket_name = ENV["S3_BUCKET_NAME"]
 
 all_question_types_form = Form.create!(
   name: "All question types form",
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -7,6 +7,8 @@
 #   Character.create(name: "Luke", movie: movies.first)
 
 submission_email = ENV["EMAIL"] || `git config --get user.email`.strip
+s3_bucket_aws_account_id = ENV["AWS_ACCOUNT_ID"]
+s3_bucket_name = ENV["S3_BUCKET_NAME"]
```
stephencdaly, line 11: People probably wouldn't set these environment variables when running the seeds normally, so in the instructions for running the tests locally we'd probably need to tell them to configure these settings on the "s3 submission test form" form. Currently that would be running a rake task.

We'd probably want to update the seeds on forms-admin too to put this form in the end to end tests group so people can see it in forms-admin.

alice-carr: I see that definitely makes sense. 
I'm still not entirely clear how the seeds are setup in the environments that the pipelines and how we much sure the forms is set up there so I'll have a look into that

stephencdaly: I'm not entirely sure that we do ever run the seeds on the environments - I think we just manually set up whatever we need to. But yeh would be worth checking that. I think the seeds are just for local development and are generally infrequently run - usually only whenever there is some data migration where it's easier to just destroy and recreate your local database. 

We could suggest that people re-run the seeds passing in those environment variables when they need to run these tests, its probably quite infrequent that people will need to run these specific tests locally.

PR #627, opened 2024-11-01
Add form with branch routing to database seed

lfdebrux:

For local developer setups, this make it easier to see what branching looks like, by adding an example form with branching to the database seed.


DavidBiddle (dismissed):
The seed data looks reasonable, and I can see it in my local database and use it successfully in runner after running `bin/rails db:seed` 🎉

DavidBiddle (approved):

PR #629, opened 2024-11-05
add new form to seeds for end to end testing

alice-carr:

we're adding a new end to end test that will test forms with a submission type of s3. There's no front end way to set a form  to a type of s3, so we can't create the form in the e2e tests. Instead, when testing locally, we will use this new seeded form.

[Trello card](https://trello.com/c/64CM50hZ/1921-add-an-e2e-test-for-s3-submissions)


lfdebrux (approved):
Works for me!

PR #637, opened 2024-11-15
Add API endpoints to generate detailed reports for usage of select from a list questions

stephencdaly:

Trello card: https://trello.com/c/SVZMGF6m/

  Add API endpoints to generate reports for:
  - A summary report with numbers of forms and questions with selection
  questions that will use each of autocomplete, radios and checkboxes
  when they're displayed in runner.
  - Individual reports to list all questions that will use each of
  autocomplete, radios and checkboxes when they're displayed in runner.

Forms-admin PR to display these reports: https://github.com/alphagov/forms-admin/pull/1604
 
### DIFF
--- a/app/controllers/api/v1/reports_controller.rb
+++ b/app/controllers/api/v1/reports_controller.rb
@@ -4,4 +4,28 @@ class Api::V1::ReportsController < ApplicationController
 
     render json: feature_stats.to_json, status: :ok
   end
+
+  def selection_questions_summary
+    statistics = Reports::SelectionQuestionService.new.live_form_statistics
+
+    render json: statistics.to_json, status: :ok
+  end
+
+  def selection_questions_with_autocomplete
+    data = Reports::SelectionQuestionService.new.live_form_pages_with_autocomplete
+
+    render json: data.to_json, status: :ok
+  end
+
+  def selection_questions_with_radios
+    data = Reports::SelectionQuestionService.new.live_form_pages_with_radios
+
+    render json: data.to_json, status: :ok
+  end
+
+  def selection_questions_with_checkboxes
+    data = Reports::SelectionQuestionService.new.live_form_pages_with_checkboxes
+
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/reports_controller.rb
```diff
@@ -4,4 +4,29 @@ def features
...
+    statistics = Reports::SelectionQuestionService.new.live_form_statistics
+
+    render json: statistics.to_json, status: :ok
+  end
+
+  def selection_questions_with_autocomplete
+    questions = Reports::SelectionQuestionService.new.live_form_pages_with_autocomplete
+
+    # adding the count as ActiveResource doesn't like parsing JSON with a single root key
+    render json: { questions:, count: questions.length }.to_json, status: :ok
```
DavidBiddle, line 18: Do you reckon it's worth returning the count in the various `Reports::SelectionQuestionService` methods?

stephencdaly: Ah so just move where we construct the JSON response with the count into the service instead? Yeh I can do that.

I think my reasoning for putting it here was that we're only adding it due to a limitation of ActiveResource rather than actually needing it. So I thought putting it in the controller was slightly closer to the request and so the comment with the reasoning makes more sense here. But also don't feel too strongly.

DavidBiddle: My thinking was that even if we're just doing it to make ActiveResource happy, it's a statistic that could make sense in the context of the SelectionQuestionService. But your logic makes sense too so I'm happy either way 😁

stephencdaly: Yeh ok, that reasoning is good enough for me - I'll move it

stephencdaly: Made the change

DavidBiddle (approved):

PR #648, opened 2024-12-06
Delete the secondary skip for a branch route if the branching question is deleted

lfdebrux:

Trello card: https://trello.com/c/UThSZ8ga/2037-fix-behaviour-of-deleting-page-with-secondary-skip-route 

It doesn't makes sense to have a secondary skip condition when the precondition skip route has been deleted, but that's currently what happens when you delete a page with a skip and secondary skip.

Instead let's just delete all check conditions for a page when that page is deleted, same as we do for the routing conditions. This shouldn't have any other side effects, as the only case currently where a check condition isn't also a routing condition for a page is for a secondary skip.


thomasiles (approved):
looks good and works 🏅

PR #649, opened 2024-12-09
Fix condition validations

thomasiles:
# Change routing validations to take into account secondary skips

Trello card: https://trello.com/c/5kRIGgDY/2581-change-api-validations-for-conditions-to-use-routingpageid-instead-of-checkpageid

The validations used to make sure conditions make sense within a form do not take into account secondary skips.

This creates issues when displaying secondary skips within the page list and can result in forms with loops being made live.

Because the exisitning validation code used the check_page as a reference it could report incorrect errors in two ways.

- It would produce errors which say that a secondary skip is invalid because the route is next to the end route, even when it wasn't
- It would miss loops in forms

Both cases are fixed by changing the validation code to use `routing_page` instead of `check_page`. For exisiting forms with a basic route, this shouldn't result in any changes.

For forms created with secondary skips, it should now report the correct errors.

### Example of an incorrect error based on position
image

### Example of a form with a loop and no error
image

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -59,14 +59,9 @@ class Condition < ApplicationRecord
   end
 
   def warning_goto_page_before_check_page
-    return nil if check_page.nil? || goto_page.nil?
-
-    routing_page_position = routing_page.position
-    goto_page_position = goto_page.position
-
-    return { name: "cannot_have_goto_page_before_routing_page" } if goto_page_position < (routing_page_position + 1)
-
-    nil
+    if goto_page.present? && (goto_page.position <= routing_page.position)
+      { name: "cannot_have_goto_page_before_routing_page" }
+    end
   end
 
   def is_check_your_answers?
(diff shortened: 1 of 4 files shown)


DavidBiddle (dismissed):
Changes are well explained, code looks good, and the behaviour on my local machine looks right 🎉

lfdebrux:
Tricky tricky tricky! Nice one spotting this and figuring it out. I'm worried the difference between routing pages and check pages is possibly going to be a source of more such subtle bugs in future 😬 

I think for this particular bug it would be good to get some unit tests to help make it obvious what the fix is about, and to prevent regressions; would that be possible do you think?

thomasiles:
> Tricky tricky tricky! Nice one spotting this and figuring it out. I'm worried the difference between routing pages and check pages is possibly going to be a source of more such subtle bugs in future 😬
> 
> I think for this particular bug it would be good to get some unit tests to help make it obvious what the fix is about, and to prevent regressions; would that be possible do you think?

Yes, it's a little spooky that all the existing unit tests pass. It's consistent in the way we build conditions, it's just that routing_page has always been the same as check_page.

I'll change the existing tests to make it clearer that secondary skips play a role in this.

Hopefully changes like cascading page deletion to conditions will reduce the number of states we have to think about for conditions but it is tricky.

DavidBiddle (approved):

PR #663, opened 2024-12-23
Add report to get all live forms

stephencdaly:

Trello card: https://trello.com/c/mPtJhtWq/

Add API endpoint to retrieve all live forms for the purpose of generating reports. This uses the "api-pagination" and "pagy" gems to paginate the results to avoid sending all the form data in one API response.

We return all the form data for forms-admin to process so that organisation and group data can be added to the report.

Related admin PR to add the report: https://github.com/alphagov/forms-admin/pull/1681


app/controllers/api/v2/form_documents_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
 
     documents = Api::V2::FormDocument.all
     documents = documents.where(tag:) if tag.present?
-    paginated = documents.page(params[:page]).per(params[:per_page] || DEFAULT_PAGE_SIZE)
+    paginated = documents.page(params[:page]).per(params[:per_page] || DEFAULT_PAGE_SIZE).order(:id)
```
DavidBiddle, line 11: I think it would make sense to update the commit name to `Order returned form documents by id` to more accurately describe what this is doing

stephencdaly: Good spot, I changed the ordering but forgot to update the message! 🤦‍♂️

stephencdaly: done

DavidBiddle (commented):
Raised a comment about one of the commit messages, but otherwise looks great and works for me locally

DavidBiddle (approved):

PR #679, opened 2025-01-27
Bump base image

chao-xian:
https://trello.com/c/Q1UuCEPI/2664-ruby-version-update-to-341

Bumps core dependencies and base image in Dockerfile.

- Bumps Ruby from 3.3.4 to 3.4.1
- Bumps Alpine Linux from 3.19 to 3.21


DavidBiddle (dismissed):

Gemfile.lock
```diff
@@ -357,7 +358,15 @@ PLATFORMS
...
   arm64-darwin
+  arm64-darwin-21
+  arm64-darwin-22
+  arm64-darwin-23
+  arm64-darwin-24
   x86_64-darwin
+  x86_64-darwin-21
+  x86_64-darwin-22
+  x86_64-darwin-23
+  x86_64-linux
```
lfdebrux, line 369: I think we don't want these lines if we can avoid it

```suggestion
  x86_64-darwin
```

lfdebrux (approved):

PR #686, opened 2025-02-06
Change condition model to delete secondary skip routes when skip route is deleted

lfdebrux:

Trello card: https://trello.com/c/iMnMgYPr/2106-deleting-a-route-should-delete-the-secondary-skip-route-for-the-same-question 

Currently deleting a skip route for a question that has a secondary skip route puts the form in a state where there is an invalid route that can’t be deleted.

We've decided to fix this by making it so that deleting a route for a page also deletes any secondary skip routes for that page.

Because routes and associations between conditions are directly expressed in our model, we need to jump through some hoops to make this work. This commit adds a before_destroy callback to take care of things.


jamie-o-wilkinson (approved):
Have talked this through with @lfdebrux, all looks good to me 👍

PR #700, opened 2025-02-24
Remove unused code and directories

theseanything:


lfdebrux:
The `tmp` and `log` folders are used when running the app, do we not need them to be kept in git? 🤔

theseanything:
> The `tmp` and `log` folders are used when running the app, do we not need them to be kept in git? 🤔

Sorry hadn't updated the description before, but Rails will create them as needed anyways.

lfdebrux (approved):
Works for me locally. Thanks!

chao-xian:
congrats @theseanything on #700 ! 🥳

PR #704, opened 2025-03-07
Add markdown for exit pages

thomasiles:
### Add columns for storing the markdown content for exit pages to Condition

Trello card: https://trello.com/c/LYs67GUv/2150-add-somewhere-to-store-exit-page-guidance-in-our-database

This PR adds two columns to the Condition model for storing the heading and markdown content for exit pages.

Exit pages will be represented by a Condition with check_page_id, routing_page_id set and a string value, which could be "", for exit_page_markown.

The Condition validation which reports an error when the goto_page is nil has been changed to ignore exit pages.

It doesn't add any other validations we might need for exit pages. The guidance content stored in pages [uses validations which might be useful](https://github.com/alphagov/forms-api/blob/ff61d5a8b84108b872b00fe074c08ad62cab1d19/app/models/page.rb#L75) such as checking the heading is set if the content is set and checks on content and length.

The branching example form in the seed has been extended to include an exit page.

We are keeping the admin and API database table definitions in sync to make migrating easier. Another PR will be raised against admin when this is approved to add the same fields to the conditions table. 

[Displaying exit pages to form fillers](https://trello.com/c/pfNYulxC/2151-add-support-for-exit-pages-to-forms-runner) and allowing form creators to create and change exit pages and separate tickets.

 
- Does this sound like a good way to represent exit pages?
- Does the design allow us creating/modifying exit pages?
- Does it make anything about our current condition code harder?
- Does it make it possible to display exit_pages to form fillers?

### DIFF
--- a/db/migrate/20250306120134_add_exit_page_fields_to_conditions.rb
+++ b/db/migrate/20250306120134_add_exit_page_fields_to_conditions.rb
@@ -1,0 +1,8 @@
+class AddExitPageFieldsToConditions < ActiveRecord::Migration[8.0]
+  def change
+    change_table(:conditions, bulk: true) do |t|
+      t.column :exit_page_markdown, :text, comment: "When not nil this condition should be treated as an exit page. When set it contains the markdown for the body of the exit page"
+      t.column :exit_page_heading, :text, comment: "Text for the heading of the exit page"
+    end
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -122,6 +122,18 @@ branch_route_form = Form.create!(
   name: "Branch route form",
   pages: [
     Page.create(
+      question_text: "Are you eligible to submit this form?",
+      answer_type: "selection",
+      answer_settings: {
+        only_one_option: "true",
+        selection_options: [
+          { "name": "Yes" },
+          { "name": "No" },
+        ],
+      },
+      is_optional: false,
+    ),
+    Page.create(
       question_text: "How many times have you filled out this form?",
       answer_type: "selection",
       answer_settings: {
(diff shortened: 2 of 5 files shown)


db/seeds.rb
```diff
@@ -167,6 +167,27 @@
...
+      answer_type: "selection",
+      answer_settings: {
+        only_one_option: "true",
+        selection_options: [
+          { "name": "Yes" },
+          { "name": "No" },
+        ],
+      },
+      is_optional: false,
+    ),
```
lfdebrux, line 181: One minor note, I think you'd normally put the exit page at the beginning of a form! 😁

thomasiles: Thats a good point, I was trying to avoid having to change all the page ids :)

db/migrate/20250306120134_add_exit_page_fields_to_conditions.rb
```diff
@@ -0,0 +1,6 @@
+class AddExitPageFieldsToConditions < ActiveRecord::Migration[8.0]
+  change_table(:conditions, bulk: true) do |t|
```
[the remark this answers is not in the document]
thomasiles: I'll take a look. I was bullied into this format by rubocop but I think there are other ways around it's complaints

db/seeds.rb
thomasiles: It was actually less work to put it at the start 🙄 

I've moved it and removed the extra question.

db/migrate/20250306120134_add_exit_page_fields_to_conditions.rb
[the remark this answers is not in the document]
thomasiles: I just needed to wrap it in `def change`

lfdebrux (approved):
Looks good to me 🎉

PR #726, opened 2025-04-04
Remove feature usage report endpoint

stephencdaly:

Trello card: https://trello.com/c/SgDCKXcY/145-improve-our-feature-reports

This report is now constructed in forms-admin using data retrieved from `api/v2/form_documents` and so this endpoint is no longer consumed.

Rename the `FeatureUsageService` to `AddAnotherAnswerUsage` service as this is the only thing it is now used for.


lfdebrux (dismissed):
Yay for deleting things!

lfdebrux:
@stephencdaly Can we remove the endpoint from `config/routes.rb`?

stephencdaly:
> @stephencdaly Can we remove the endpoint from `config/routes.rb`?

Ah good spot, thanks! Have removed.

lfdebrux (approved):

PR #768, opened 2025-05-27
Synchronise form documents for update submission type tasks

stephencdaly:

forms-runner now uses the V2 API which consumes FormDocuments. After updating the forms in these rake tasks, synchronise the forms so that the FormDocuments are updated.

Remove the rake task to set submission type to email_with_csv as this is now configurable in forms-admin so we should never need to run this task.


alice-carr (approved):

thomasiles (approved):
looks good 🥇

PR #769, opened 2025-05-28
Make "db/" into a volume inside the Docker container

AP-Hunt:

Trello card: https://trello.com/c/KoHvaEUA/681-aws-m112-ecs-read-only-root-filesystem-configuration

AWS ECS docs [1] say that a file that exists in a directory that is subsequently marked as a volume will be copied to the host and remounted into the container.

This is necessary for us to do with the "db/" directory because the "db:migrate" Rake task writes to "db/schema.rb"

[1] https://docs.aws.amazon.com/AmazonECS/latest/developerguide/bind-mounts.html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,9 +58,10 @@ RUN chmod 0755 bin/*
 COPY --chown=ruby:ruby --from=build /usr/local/bundle /usr/local/bundle
 COPY --chown=ruby:ruby --from=build /app /app
 
-RUN mkdir -p "/app/tmp/" && chown ruby:ruby "/app/tmp/"
+RUN mkdir -p "/app/tmp/" && chown ruby:ruby "/app/tmp/" && chown ruby:ruby "/app/db/"
 VOLUME "/tmp/"
 VOLUME "/app/tmp/"
+VOLUME "/app/db/"
 
 EXPOSE 9292
 


AP-Hunt:
> only if there are any changes

That is true. We saw this on `forms-runner` and it caused an error. We're porting the fix over here to stop that happening.

lfdebrux:
> > only if there are any changes
> 
> That is true. We saw this on `forms-runner` and it caused an error. We're porting the fix over here to stop that happening.

Hmm. Ideally there shouldn't be changes to `db/schema.rb` from running `db:migrate`, as that indicates a configuration drift.

lfdebrux:
> > Hmm. Ideally there shouldn't be changes to db/schema.rb from running db:migrate, as that indicates a configuration drift. Ideally any changes would have been committed to the main branch before the deploy.
> 
> That's .. concerning? Probably? We were seeing this in staging, and in review once we began running `db:migrate` and not just `db:setup`: [alphagov/forms-runner@7d943ce](https://github.com/alphagov/forms-runner/commit/7d943ce3faa39c933a85aef248204c4e04b6f36d)

Yeah I saw that. It is concerning, but only slightly? Probably not something that needs to be fixed right this moment.

PR #791, opened 2025-07-09
Prepare next_page column in pages table to be removed

lfdebrux:

While working on migrating the forms models from forms-api to forms-admin I noticed that the next_page column for a page in the forms-api database is always set to nil.

As far as I can tell this column hasn't been used since before we migrated from grape (see PR #116), when we migrated to grape we switched to using the acts_as_list gem and the position column instead (see commit e8cc03).

I think we should remove this column now.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -77,6 +77,7 @@ class Form < ApplicationRecord
                 include: {
                   routing_conditions: { methods: :validation_errors },
                 },
+                methods: [:next_page],
               },
             }, methods: [:start_page]).merge(kwargs)
   end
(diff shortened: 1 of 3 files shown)


theseanything (dismissed):

lfdebrux:
> @lfdebrux how come we're adding in tests that expect next_page to not be nil?

So it's a bit confusing, but we do still rely on the `next_page` attribute in forms-runner and forms-admin, but it is generated by a method in https://github.com/alphagov/forms-api/blob/main/app/models/page.rb#L47-L49 that uses the `position` attribute, rather than coming from the database. So we don't need the `next_page` column, but we do need the `next_page` attribute in the JSON responses.

theseanything (approved):

PR #794, opened 2025-07-11
wip

SamJamCul:

Trello card: https://trello.com/c/5WfLpFRG/2373-move-feature-report-code-from-forms-api-to-forms-admin

Once https://github.com/alphagov/forms-admin/pull/2072 is in, we can remove all the migrated reports from Forms API. This PR gets rid of the report services and the endpoints that served them. 


lfdebrux (commented):
Oh, we can also remove the routes from `config/routes.rb`

